PR #87, opened 2012-01-15
Added clear-tube command to beanstalkd 

julman99:
The clear-tube command WILL clear the ready and the delayed jobs from the specified tube.

Syntax:
clear-tube <tube_name>

The tube will be immediately marked as empty, all the jobs will be copied to a trash. On each tick, the server will free the memory of 5 jobs from the trash list. Note that the number 5 was chosen arbitrarily and probably there are better values which will free the memory faster without performance hit.


contentfree:
Any additional movement here? This would be a very useful command on occasion

JensRantil:
Before this is being implemented I think it's important to ask; What is stopping you from simply doing

```
while beanstalk.stats.jobs > 0:
    job = beanstalk.reserve(timeout=0)
    if job is None:
        break
    job.delete()
```

It will essentially drain beanstalk and will also make sure that no other consumer/worker has any job reserved.

contentfree:
Yes, it just gets tedious and when there's a few hundred thousand jobs, it's
(relatively) slow. Having a clear-tube command just seems like something that
should exist.

PR #150, opened 2012-10-20
AF_UNIX support for beanstalkd

etanol:
Add the posibility of accepting connections from a local socket (bound to a filesystem path).

To minimize portability issues, socket paths longer than 100 bytes are rejected.  Systemd support has been added too, but not tested as I don't have access to any Fedora or similar.

The behaviour of the new command line switch may be strange, but it has been implemented in a way to avoid script breakage when this option is not used.

The automatic socket file replacement may be a bit surprising too.  However, I found that implementing an automatic removal/cleanup on exit was a bit more intrusive.

Any feedback will be appreciated.

### DIFF
--- a/doc/beanstalkd.1
+++ b/doc/beanstalkd.1
@@ -1,7 +1,7 @@
 .\" generated with Ronn/v0.7.3
 .\" http://github.com/rtomayko/ronn/tree/0.7.3
 .
-.TH "BEANSTALKD" "1" "April 2012" "" ""
+.TH "BEANSTALKD" "1" "August 2019" "" ""
 .
 .SH "NAME"
 \fBbeanstalkd\fR \- simple, fast work queue
@@ -55,6 +55,9 @@ Show a brief help message and exit\.
 .TP
 \fB\-l\fR \fIaddr\fR
 Listen on address \fIaddr\fR (default is 0\.0\.0\.0)\.
+.
+.IP
+When \fIaddr\fR starts with "unix:", the unprefixed value of it will be used as the local filesystem path to create a UNIX socket instead of a TCP socket\. In this case the value of \fB\-p\fR will be ignored\.
 .
 .IP
 (Option \fB\-l\fR has no effect if sd\-daemon(5) socket activation is being used\. See also \fIENVIRONMENT\fR\.)
--- a/doc/beanstalkd.1.html
+++ b/doc/beanstalkd.1.html
@@ -121,6 +121,10 @@ it writes to the binlog.</p>
 <dt class="flush"><code>-h</code></dt><dd><p>Show a brief help message and exit.</p></dd>
 <dt class="flush"><code>-l</code> <var>addr</var></dt><dd><p>Listen on address <var>addr</var> (default is 0.0.0.0).</p>
 
+<p>When <var>addr</var> starts with "unix:", the unprefixed value of it will be
+used as the local filesystem path to create a UNIX socket instead of
+a TCP socket. In this case the value of <code>-p</code> will be ignored.</p>
+
 <p>(Option <code>-l</code> has no effect if <span class="man-ref">sd-daemon<span class="s">(5)</span></span> socket activation is
 being used. See also <a href="#ENVIRONMENT" title="ENVIRONMENT" data-bare-link="true">ENVIRONMENT</a>.)</p></dd>
 <dt class="flush"><code>-n</code></dt><dd><p>Turn off binlog compaction, negating <code>-c</code>.</p>
@@ -167,7 +171,7 @@ of many others.</p>
 
   <ol class='man-decor man-foot man foot'>
     <li class='tl'></li>
-    <li class='tc'>April 2012</li>
+    <li class='tc'>August 2019</li>
     <li class='tr'>beanstalkd(1)</li>
   </ol>
 
--- a/doc/beanstalkd.ronn
+++ b/doc/beanstalkd.ronn
@@ -59,6 +59,10 @@ and format of the `beanstalkd` protocol.
 * `-l` <addr>:
   Listen on address <addr> (default is 0.0.0.0).
 
+  When <addr> starts with "unix:", the unprefixed value of it will be
+  used as the local filesystem path to create a UNIX socket instead of
+  a TCP socket. In this case the value of `-p` will be ignored.
+
   (Option `-l` has no effect if sd-daemon(5) socket activation is
   being used. See also [ENVIRONMENT][].)
 
--- a/net.c
+++ b/net.c
@@ -7,19 +7,205 @@
 #include <string.h>
 #include <errno.h>
 #include <sys/socket.h>
+#include <sys/un.h>
+#include <sys/stat.h>
 #include <netinet/in.h>
 #include <netinet/tcp.h>
 
-int
-make_server_socket(char *host, char *port)
+static int
+set_nonblocking(int fd)
+{
+    int flags, r;
+
+    flags = fcntl(fd, F_GETFL, 0);
+    if (flags < 0) {
+        twarn("getting flags");
+        return -1;
+    }
+    r = fcntl(fd, F_SETFL, flags | O_NONBLOCK);
+    if (r == -1) {
+        twarn("setting O_NONBLOCK");
+        return -1;
+    }
+    return 0;
+}
+
+static int
+make_inet_socket(char *host, char *port)
 {
     int fd = -1, flags, r;
     struct linger linger = {0, 0};
     struct addrinfo *airoot, *ai, hints;
 
+    memset(&hints, 0, sizeof(hints));
+    hints.ai_family = PF_UNSPEC;
+    hints.ai_socktype = SOCK_STREAM;
+    hints.ai_flags = AI_PASSIVE;
+    r = getaddrinfo(host, port, &hints, &airoot);
+    if (r != 0) {
+        twarnx("getaddrinfo(): %s", gai_strerror(r));
+        return -1;
+    }
+
+    for (ai = airoot; ai; ai = ai->ai_next) {
+        fd = socket(ai->ai_family, ai->ai_socktype, ai->ai_protocol);
+        if (fd == -1) {
+            twarn("socket()");
+            continue;
+        }
+
+        r = set_nonblocking(fd);
+        if (r == -1) {
+            close(fd);
+            continue;
+        }
+
+        flags = 1;
+        r = setsockopt(fd, SOL_SOCKET, SO_REUSEADDR, &flags, sizeof flags);
+        if (r == -1) {
+            twarn("setting SO_REUSEADDR on fd %d", fd);
+            close(fd);
+            continue;
+        }
+        r = setsockopt(fd, SOL_SOCKET, SO_KEEPALIVE, &flags, sizeof flags);
+        if (r == -1) {
+            twarn("setting SO_KEEPALIVE on fd %d", fd);
+            close(fd);
+            continue;
+        }
+        r = setsockopt(fd, SOL_SOCKET, SO_LINGER, &linger, sizeof linger);
+        if (r == -1) {
+            twarn("setting SO_LINGER on fd %d", fd);
+            close(fd);
+            continue;
+        }
+        r = setsockopt(fd, IPPROTO_TCP, TCP_NODELAY, &flags, sizeof flags);
+        if (r == -1) {
+            twarn("setting TCP_NODELAY on fd %d", fd);
+            close(fd);
+            continue;
+        }
+
+        r = bind(fd, ai->ai_addr, ai->ai_addrlen);
+        if (r == -1) {
+            twarn("bind()");
+            close(fd);
+            continue;
+        }
+        if (verbose) {
+            char hbuf[NI_MAXHOST], pbuf[NI_MAXSERV], *h = host, *p = port;
+            struct sockaddr_in addr;
+            socklen_t addrlen;
+
+            addrlen = sizeof(addr);
+            r = getsockname(fd, (struct sockaddr *) &addr, &addrlen);
+            if (!r) {
+                r = getnameinfo((struct sockaddr *) &addr, addrlen,
+                                hbuf, sizeof(hbuf),
+                                pbuf, sizeof(pbuf),
+                                NI_NUMERICHOST|NI_NUMERICSERV);
+                if (!r) {
+                    h = hbuf;
+                    p = pbuf;
+                }
+            }
+            if (ai->ai_family == AF_INET6) {
+                printf("bind %d [%s]:%s\n", fd, h, p);
+            } else {
+                printf("bind %d %s:%s\n", fd, h, p);
+            }
+        }
+
+        r = listen(fd, 1024);
+        if (r == -1) {
+            twarn("listen()");
+            close(fd);
+            continue;
+        }
+
+        break;
+    }
+
+    freeaddrinfo(airoot);
+
+    if(ai == NULL)
+        fd = -1;
+
+    return fd;
+}
+
+static int
+make_unix_socket(char *path)
+{
+    int fd = -1, r;
+    struct stat st;
+    struct sockaddr_un addr;
+    const size_t maxlen = sizeof(addr.sun_path) - 1; // Reserve the last position for '\0'
+
+    memset(&addr, 0, sizeof(struct sockaddr_un));
+    addr.sun_family = AF_UNIX;
+    if (strlen(path) > maxlen) {
+        warnx("socket path %s is too long (%ld characters), where maximum allowed is %ld",
+              path, strlen(path), maxlen);
+        return -1;
+    }
+    strncpy(addr.sun_path, path, maxlen);
+
+    r = stat(path, &st);
+    if (r == 0) {
+        if (S_ISSOCK(st.st_mode)) {
+            warnx("removing existing local socket to replace it");
+            r = unlink(path);
+            if (r == -1) {
+                twarn("unlink");
+                return -1;
+            }
+        } else {
+            twarnx("another file already exists in the given path");
+            return -1;
+        }
+    }
+
+    fd = socket(AF_UNIX, SOCK_STREAM, 0);
+    if (fd == -1) {
+        twarn("socket()");
+        return -1;
+    }
+
+    r = set_nonblocking(fd);
+    if (r == -1) {
+        close(fd);
+        return -1;
+    }
+
+    r = bind(fd, (struct sockaddr *) &addr, sizeof(struct sockaddr_un));
+    if (r == -1) {
+        twarn("bind()");
+        close(fd);
+        return -1;
+    }
+    if (verbose) {
+        printf("bind %d %s\n", fd, path);
+    }
+
+    r = listen(fd, 1024);
+    if (r == -1) {
+        twarn("listen()");
+        close(fd);
+        return -1;
+    }
+
+    return fd;
+}
+
+int
+make_server_socket(char *host, char *port)
+{
+    int fd = -1, r;
+
     /* See if we got a listen fd from systemd. If so, all socket options etc
-     * are already set, so we check that the fd is a TCP listen socket and
-     * return. */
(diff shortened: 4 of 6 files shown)


ShadowOfHarbringer:
I would be greatly interested in this feature. 
Some of my servers handle so many connections that i already had problems with too many open ports limit on Linux.

petri:
@kr, I also have come to desire this feature. Any chance you could merge it and release? The beanstalkc Python client apparently has already support for unix socket, just waiting for beanstalkd itself to support it, and I would gladly add support to pybeanstalk as well.

silviucpp:
Hello @etanol can you please rebase the changes with the last master ?

Silviu

djzort:
Any update on this?

etanol:
I believe this project is about to transition into *abandoned* status.

sshilko:
feels bad, but you are correct, there are many fresh alternatives (NSQ/NATS/....) nsq has very similar API but i liked beanstalkd for stability

doc/beanstalkd.1
```diff
@@ -57,6 +57,9 @@ Show a brief help message and exit\.
 Listen on address \fIaddr\fR (default is 0\.0\.0\.0)\.
 .
 .IP
+When \fIaddr\fR starts with "unix:", the unprefixed value of will be used as the local filesystem path to create a UNIX socket instead of a TCP socket\. In this case the vale of \fB\-p\fR will be ignored\.
```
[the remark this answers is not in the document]
etanol: How can I fix that from the *ronn* file?  Because this ROFF code was autogenerated.

meox:
this is a great PR please rework on it

ysmolsky:
@etanol can you please look at it again?

doc/beanstalkd.1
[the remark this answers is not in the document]
pprindeville: okay, never mind... the preprocessor is broken.

net.c
```diff
@@ -37,104 +215,23 @@ make_server_socket(char *host, char *port)
...
-      r = listen(fd, 1024);
-      if (r == -1) {
-        twarn("listen()");
-        close(fd);
-        continue;
-      }
-
-      break;
+    if (host && strlen(host) > 5 && !strncmp(host, "unix:", 5)) {
+        return make_unix_socket(host + 5);
```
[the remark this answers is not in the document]
pprindeville: Actually, you could drop the `strlen(host) > 5` altogether... `strncmp()` will do the right thing in all cases.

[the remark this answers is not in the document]
etanol: You think?  What if the length of the `host` string is less than 5?  Wouldn't `strncmp` read past the boundaries of `host`?

Fudoshiki:
Any plan to merge this?

etanol:
I will give it one more try.  But this is getting ridiculous, seven years...

Fudoshiki:
This is very useful feature

ysmolsky:
@etanol thanks for doing this. many things has changed since that time, so thanks again.

ysmolsky:
@etanol, thank you. Do you plan to address the review comments by @pprindeville? Do you have a patience for my review comments as well?

etanol:
@ysmolsky Sorry for that, I forgot about those amendments.  I have updated the PR with the updates.  I also added some extra tests to keep coverage on check.

net.c
```diff
@@ -7,19 +7,197 @@
...
+        return -1;
+    }
+
+    if (verbose) {
+        printf("bind %s\n", path);
+    }
+    addr.sun_family = AF_UNIX;
+    if (strlen(path) > maxlen) {
+        warnx("truncating local socket path to: %.*s", (int)maxlen, path);
+        strncpy(addr.sun_path, path, maxlen);
```
pprindeville: What about using strlcpy() instead?

testserv.c
```diff
@@ -141,30 +165,39 @@ kill_srvpid(void)
...
-mustforksrv(void)
+mustforksrv(char *socket_file)
 {
-    struct sockaddr_in addr;
+    int port = -1;
 
-    srv.sock.fd = make_server_socket("127.0.0.1", "0");
+    if (socket_file != NULL) {
+        char name[200];
+        snprintf(name, 200, "unix:%s", socket_file);
```
pprindeville: sizeof(name) instead of 200...

testserv.c
```diff
@@ -343,6 +376,20 @@ cttest_too_long_commandline()
...
 
+void
+cttest_too_long_socketname()
+{
+    #define NAME "/tmp/aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+    progname = __func__;
+    struct sockaddr_un addr;
+    const size_t maxlen = sizeof(addr.sun_path);
+    char real_name[maxlen];
+    snprintf(real_name, maxlen, NAME);
```
pprindeville: There's no %-format specifier, so please don't use snprintf(); use strlcpy() instead.

testserv.c
```diff
@@ -383,6 +430,41 @@ cttest_peek_not_found()
     ckresp(fd, "NOT_FOUND\r\n");
 }
 
+void
+cttest_peek_ok_unix()
+{
+    progname = __func__;
+    char name[90];
+    snprintf(name, 90, "/tmp/ctb-%d", getpid());
```
pprindeville: sizeof(name) instead of 90...

testserv.c
```diff
@@ -383,6 +430,41 @@ cttest_peek_not_found()
...
+
+    unlink(name);
+}
+
+void
+cttest_unix_auto_removal()
+{
+    progname = __func__;
+    char name[90];
+    snprintf(name, 90, "/tmp/ctb-%d", getpid());
```
pprindeville: sizeof(name)

net.c
etanol: It's not as standard, and it would require quite a bit of boilerplate to compile it properly under the various UNIX flavours.

testserv.c
etanol: Good point, although I'll stick to `snprintf`.  Remember that this is just test code, so the fact that it works triumphs over its elegance.

net.c
```diff
@@ -7,19 +7,197 @@
...
+        addr.sun_path[maxlen] = '\0';
+    } else {
+        strcpy(addr.sun_path, path);
+    }
```
ysmolsky: There is no need to do copying in the branches. You can leave the warning in the branch and just copy string afterwards. It wont copy more than len of `path` and it will fill the rest of the `addr.sun_path` buffer with zeroes:
```c
    if (strlen(path) > maxlen) {
        warnx("truncating local socket path to: %.*s", (int)maxlen, path);
    }
    strncpy(addr.sun_path, path, maxlen);
```

net.c
```diff
@@ -7,19 +7,197 @@
...
+    r = set_nonblocking(fd);
+    if (r == -1) {
+        close(fd);
+        return -1;
+    }
+
+    if (verbose) {
+        printf("bind %s\n", path);
+    }
+    addr.sun_family = AF_UNIX;
```
ysmolsky: Please add `memset(&addr, 0, sizeof(struct sockaddr_un addr));` before using the structure.

net.c
ysmolsky: There is no need for strlcpy() here, see below.

net.c
```diff
@@ -32,118 +210,28 @@ make_server_socket(char *host, char *port)
         fd = SD_LISTEN_FDS_START;
         r = sd_is_socket_inet(fd, 0, SOCK_STREAM, 1, 0);
         if (r < 0) {
-            errno = -r;
             twarn("sd_is_socket_inet");
+            errno = -r;
             return -1;
         }
         if (!r) {
```
ysmolsky: Are sure that r==0 here in the case when systemd provided listen fd for the unix socket?

etanol: [From `libsystemd`'s man pages][1]:

> ### Return value
>
> On failure, these calls return a negative errno-style error code. If the file descriptor is of the specified type and bound to the specified address, a positive return value is returned, otherwise zero.

[1]: https://www.freedesktop.org/software/systemd/man/sd_is_socket_inet.html#

net.c
```diff
@@ -7,19 +7,197 @@
...
+        r = setsockopt(fd, IPPROTO_TCP, TCP_NODELAY, &flags, sizeof flags);
+        if (r == -1) {
+            twarn("setting TCP_NODELAY on fd %d", fd);
+            close(fd);
+            continue;
+        }
+
+        if (verbose) {
+            char hbuf[NI_MAXHOST], pbuf[NI_MAXSERV], *h = host, *p = port;
+            r = getnameinfo(ai->ai_addr, ai->ai_addrlen,
```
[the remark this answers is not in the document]
etanol: Accident, my bad.

net.c
ysmolsky: Ah, so it is the same as with case for inet socket above. Sounds right then.

net.c
```diff
@@ -7,143 +7,237 @@
...
+        }
+        if (verbose) {
+            char hbuf[NI_MAXHOST], pbuf[NI_MAXSERV], *h = host, *p = port;
+            struct sockaddr_in addr;
+            socklen_t addrlen;
+
+            addrlen = sizeof(addr);
+            r = getsockname(fd, (struct sockaddr *) &addr, &addrlen);
+            if (!r) {
+                r = getnameinfo(ai->ai_addr, ai->ai_addrlen,
```
ysmolsky: Original line is `             r = getnameinfo((struct sockaddr *) &addr, addrlen,`, your version looks too complicated. What is purpose of writing into `ai->ai_addr` instead of local var `addr`?

net.c
```diff
@@ -7,143 +7,237 @@
...
-      r = bind(fd, ai->ai_addr, ai->ai_addrlen);
-      if (r == -1) {
+    memset(&addr, 0, sizeof(struct sockaddr_un));
+    addr.sun_family = AF_UNIX;
+    if (strlen(path) > maxlen) {
+        warnx("truncating local socket path to: %.*s", (int)maxlen, path);
+    }
+    strncpy(addr.sun_path, path, maxlen);
+
+    r = bind(fd, (struct sockaddr *) &addr, sizeof(struct sockaddr_un));
```
ysmolsky: I guess you meant `struct sockaddr_un` in both of casts.

net.c
```diff
@@ -7,143 +7,237 @@
...
-      }
+        return -1;
+    }
 
-      r = bind(fd, ai->ai_addr, ai->ai_addrlen);
-      if (r == -1) {
+    memset(&addr, 0, sizeof(struct sockaddr_un));
+    addr.sun_family = AF_UNIX;
+    if (strlen(path) > maxlen) {
+        warnx("truncating local socket path to: %.*s", (int)maxlen, path);
```
ysmolsky: I propose that we `return -1` in case of truncated socket path. We should fail clearly ASAP of socket path is not correct. The test looked fishy, so I examined behaviour here, we should not allow this kind of error because when path was truncated then user definitely has an issue. We can keep the `strncpy` as safety measure though.

And test can check the returned results to check that long socket path or path to wrong directory won't be tolerated by the program.

testserv.c
```diff
@@ -383,6 +430,41 @@ cttest_peek_not_found()
...
 }
 
+void
+cttest_peek_ok_unix()
+{
+    progname = __func__;
+    char name[90];
+    snprintf(name, sizeof(name), "/tmp/ctb-%d", getpid());
+    mustforksrv(name);
+    assert(exist(name));
```
ysmolsky: These 5 lines above should be simplified. Please look at the ctdir() function. This functions returns path to the temp directory, so there is no need to do stuff with getpid(). I think that name for the socket path should be generated in the forking function.

testserv.c
```diff
@@ -343,6 +376,20 @@ cttest_too_long_commandline()
...
+void
+cttest_too_long_socketname()
+{
+    #define NAME "/tmp/aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+    progname = __func__;
+    struct sockaddr_un addr;
+    const size_t maxlen = sizeof(addr.sun_path);
+    char real_name[maxlen];
+    snprintf(real_name, maxlen, "%s", NAME);
+    mustforksrv(NAME);
```
ysmolsky: You could call `make_server_socket()` here...

testserv.c
```diff
@@ -141,30 +165,39 @@ kill_srvpid(void)
...
 
-#define SERVER() (progname=__func__, mustforksrv())
+#define SERVER() (progname=__func__, mustforksrv(NULL))
 
 // Forks the server storing the pid in srvpid.
 // The parent process returns port assigned.
 // The child process serves until the SIGTERM is received by it.
 static int
-mustforksrv(void)
+mustforksrv(char *socket_file)
```
ysmolsky: I would prefer not to mix UNIX socket into this function. This function returns an int, port value. Your change to this function looks confusing. I suggest to copy this function to mustforksrv_unix(). This function should return a socket path. And define macro SERVER_UNIX() similar to SERVER().
It is okay to have dup code in those functions for now. Simplicity is more important than elegance or cleverness.

net.c
etanol: Again, an accident.  I never had the intention to modify the original code for the listening socket setup.  I trusted the automatic rebase and Emacs three way merge more than I should have.

net.c
etanol: Totally agree.  Even with a warning, truncating the name may just lead to other unrelated and puzzling failures.

I didn't think that thoroughly before.

net.c
etanol: No, the cast is intentional (otherwise it wouldn't be necessary), see [bind(2)'s man page][1].

- Simple answer: it wouldn't compile otherwise, because warnings are treated as errors
- Longer answer: this is C-style poor man's polymorphism from the 80's, and also why the structure length needs to be provided too

[1]: https://manpage.me/index.cgi?q=bind&sektion=2

testserv.c
etanol: Sure.  But I'm blocked by #541.  So I won't be able to update this PR until #541 is merged.

testserv.c
etanol: I initially wanted to to use that temporary directory, but I didn't search enough to figure out the `ctdir()` function was what I needed.

testserv.c
ysmolsky: Ok, let's wait for it. I won't plan any other merges too to not delay it further...

ysmolsky: I have merged #541.

ysmolsky:
I really wish that Github notified me when there is a new push into the PR. Reviewing now.

testserv.c
```diff
@@ -166,16 +196,43 @@ mustforksrv(void)
 
     int port = ntohs(addr.sin_port);
     srvpid = fork();
-    if (srvpid < 0) {
-        twarn("fork");
+    if (srvpid > 0) {
```
ysmolsky: Here the original code contained the check for srvpid < 0:
```
if (srvpid < 0) {
   twarn("fork");	
   exit(1);
}
```
The same check should happen in mustforksrv_unix

testserv.c
```diff
@@ -142,6 +166,12 @@ kill_srvpid(void)
 }
 
 #define SERVER() (progname=__func__, mustforksrv())
+#define SERVER_UNIX() ({                        \
+            progname=__func__;                  \
+            char *path = mustforksrv_unix();    \
+            assert(exist(path));                \
```
ysmolsky: Let's simplify this. We can move this assert into mustforksrv_unit() and then it can be made look similar to SERVER() macro.

testserv.c
```diff
@@ -166,16 +196,43 @@ mustforksrv(void)
...
     }
 
+    srvpid = fork();
     if (srvpid > 0) {
         // On exit the parent (test) sends SIGTERM to the child.
         atexit(kill_srvpid);
-        printf("start server port=%d pid=%d\n", port, srvpid);
-        return port;
+        printf("start server socket=%s\n", path);
+        return path;
```
ysmolsky: We could put assert after the line 234

testserv.c
```diff
@@ -363,6 +420,30 @@ cttest_peek_not_found()
     ckresp(fd, "NOT_FOUND\r\n");
 }
 
+void
+cttest_peek_ok_unix()
+{
+    char *name = SERVER_UNIX();
```
ysmolsky: Now it reads much better.

testserv.c
```diff
@@ -363,6 +420,30 @@ cttest_peek_not_found()
...
+
+    unlink(name);
+}
+
+void
+cttest_unix_auto_removal()
+{
+    // Twice, to trigger autoremoval
+    SERVER_UNIX();
+    SERVER_UNIX();
```
ysmolsky: Do we expect that previous server instance should be still running here? If not we could kill it with kill_srvpid();

testserv.c
etanol: Oops :-(

testserv.c
etanol: Not really.  So yes, killing the server in between makes sense.

doc/beanstalkd.1
ysmolsky: value of **it** will

doc/beanstalkd.1.html
```diff
@@ -121,6 +121,10 @@ <h2 id="OPTIONS">OPTIONS</h2>
 <dt class="flush"><code>-h</code></dt><dd><p>Show a brief help message and exit.</p></dd>
 <dt class="flush"><code>-l</code> <var>addr</var></dt><dd><p>Listen on address <var>addr</var> (default is 0.0.0.0).</p>
 
+<p>When <var>addr</var> starts with "unix:", the unprefixed value of will be
```
ysmolsky: value of **it** will

doc/beanstalkd.ronn
```diff
@@ -59,6 +59,10 @@ and format of the `beanstalkd` protocol.
 * `-l` <addr>:
   Listen on address <addr> (default is 0.0.0.0).
 
+  When <addr> starts with "unix:", the unprefixed value of will be
```
ysmolsky: value of **it** will

doc/beanstalkd.ronn
```diff
@@ -59,6 +59,10 @@ and format of the `beanstalkd` protocol.
...
+  used as the local filesystem path to create a UNIX socket instead of
+  a TCP socket. In this case the vale of `-p` will be ignored.
```
ysmolsky: vale -> value

doc/beanstalkd.1.html
```diff
@@ -121,6 +121,10 @@ <h2 id="OPTIONS">OPTIONS</h2>
...
+used as the local filesystem path to create a UNIX socket instead of
+a TCP socket. In this case the vale of <code>-p</code> will be ignored.</p>
```
ysmolsky: vale -> value

doc/beanstalkd.1
ysmolsky: vale -> value

etanol:
No problem, glad to help and also get some closure.

ysmolsky:
Happy merged it. 

🎉

Fudoshiki:
hmm, how use that?

petri:
> hmm, how use that?

From the docs: 

`When addr starts with "unix:", the unprefixed value of it will be used as the local filesystem path to create a UNIX socket instead of a TCP socket.`

askurihin:
Can you please make a release, that will include this feature?

ysmolsky:
Yes, I am planning release very soon.

PR #175, opened 2013-03-02
systemd updates

jney:


kr:
I was under the impression that port activation would take care of startup, so the WantedBy and After lines are not what we want. Was I wrong?

adm/systemd/beanstalkd.service
```diff
@@ -3,4 +3,4 @@ Description=Beanstalkd is a simple, fast work queue
 
 [Service]
 User=nobody
-ExecStart=/usr/bin/beanstalkd
+ExecStart=/usr/local/bin/beanstalkd
```
tlstid: 5029159663

adm/systemd/beanstalkd.service
```diff
@@ -1,6 +1,10 @@
 [Unit]
 Description=Beanstalkd is a simple, fast work queue
+After=network.target
 
 [Service]
 User=nobody
-ExecStart=/usr/bin/beanstalkd
+ExecStart=/usr/local/bin/beanstalkd
+
+[Install]
```
tlstid: 5029159663

adm/systemd/beanstalkd.service
```diff
@@ -1,6 +1,10 @@
...
+WantedBy=multi-user.target
```
tlstid: 5029159663

adm/systemd/beanstalkd.service
pprindeville: Why does the path need to change? And why not used the $(PREFIX) in the Makefile, and add a `sed` command?

I.e. you could change this line to:

```
ExecStart=@PREFIX@/beanstalkd
```

and then have in the Makefile:

```
sed -i -e 's/@PREFIX@/$(PREFIX)/g' ...
```

when the file gets installed.

jney:
i made this PR while i was on archlinux. which explain the path.
i don't use beanstalkd anymore, so i'm not relevant about systemd support. my opinion is, this PR should be closed.

adm/systemd/beanstalkd.service
pprindeville: Eh?  Is that a phone number?

JensRantil:
Hello! #393 proposed we move systems stuff out of this repository. Please chime in if you have any opinion.

ysmolsky:
To be closed.

ysmolsky:
Closing as the initial author has rejected this PR.

PR #178, opened 2013-04-24
Add System V init script for beanstalkd

santisaez:
Since beanstalkd 1.6 `beanstalkd.init` file was dropped, this script is useful for Red Hat, Debian, etc. and people building RPMs.

This is a shell script in charge of starting and stopping beanstalkd service on System V, compatible with beanstalkd 1.9, verified that works with CentOS 4/5/6. Basically is the same script that was distributed with version 1.5 modified to work with the 1.9 (`-d` daemonize option has been removed, etc.).

### DIFF
--- a/adm/systemv/beanstalkd.init
+++ b/adm/systemv/beanstalkd.init
@@ -1,0 +1,139 @@
+#!/bin/sh
+#
+# System V init script in charge of starting/stopping beanstalkd
+#
+# chkconfig: - 57 47
+# description: beanstalkd is a simple, fast work queue
+# processname: beanstalkd
+# config: /etc/sysconfig/beanstalkd
+# pidfile: /var/run/beanstalkd/beanstalkd.pid
+
+### BEGIN INIT INFO
+# Provides: beanstalkd
+# Required-Start: $local_fs $network $remote_fs
+# Required-Stop: $local_fs $network $remote_fs
+# Default-Stop: 0 1 2 6
+# Short-Description: start and stop beanstalkd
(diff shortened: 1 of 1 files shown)


kr:
Is this still an issue? Does everyone use systemd these days? These awful shell scripts are not a good way to start a service like beanstalkd. If you use "System V" init and you want it to supervise the process, there are instructions in adm/Readme.

JensRantil:
Hello! #393 proposed we move systems stuff out of this repository. Please chime in if you have any opinion.

pprindeville:
> Please chime in if you have any opinion.

I think the startup packaging is always best done at the source.

adm/systemv/beanstalkd.init
```diff
@@ -0,0 +1,139 @@
...
+
+lockfile=/var/lock/subsys/${prog}
+
+start() {
+	[ -x $exec ] || exit 5
+	echo -n $"Starting $prog: "
+
+	options="-l ${BEANSTALKD_ADDR} -p ${BEANSTALKD_PORT} -u ${BEANSTALKD_USER}"
+
+	if [ "${BEANSTALKD_MAX_JOB_SIZE}" != ""  ]; then
```
pprindeville: I would have preferred:

```
if [ -n "$BEANSTALKD_MAX_JOB_SIZE" ]; then
```

instead...

ysmolsky: We still can fix it. Feel free to make a PR. I am not very proficient in bash, truth to be told. My level is only average.

PR #180, opened 2013-04-27
Implementation of 'put-unique' command.

netsweng:
This command can be used to solve many of the de-duplication issues
by not allowing duplicate messages in the queue to begin with. It has
no impact on the performance on the regular put command as it is
implmented as a parrallel command. Instead of using the job number
or name to try and find wither a message exists already, this command
compares the entire message body to determine uniqueness.

This code is in production using a 1.4.6 version of beanstalkd. This
is a quick forward port, which has not been thoroughly tested, but the
porting was straightforward, so there probably won't be any suprises.


jithujose:
This is really a useful feature to avoid job duplications.
I can't understand why it's not in the master yet.

job.c
```diff
@@ -89,6 +89,27 @@ job_find(uint64 job_id)
     return jh;
 }
 
+job
+job_find_by_body(job j)
```
JensRantil: This function is `O(n)` where `n` is number of jobs in queue, right? Unfortunately, that's not acceptable. We want at least `O(log(n))` or better.

netsweng:
Mode 3 does not always work that way. I have a single queue, multi
server implementation to maximize throughput. Each server work
independently, and there is no downstream database to maintain state.
Our design avoid having to maintain a lot of state everywhere, which
keeps things simpler. We use Put-Unique to avoid causing the queue to
grow with multiple copies of the same request. The front end, which
decides what work is needed, can simply produce the list, and allow put-
unique to filter out anything that is already in the queue. Once a job
completes, and the artifacts put in place, then the front end will no
longer try to place that item in the queue again.
Maintaining an additional date base adds a lot, and really, forces us
into a design where we could just use the database instead of
beanstalkd.
--
  Stuart Anderson
  anderson@netsweng.com



On Sun, Aug 19, 2018, at 7:32 AM, Jens Rantil wrote:
> *@JensRantil* requested changes on this pull request.


> Hello!


>> This command can be used to solve many of the de-duplication issues> Please expand. I think it's crucial that we actually define the
> problems we are trying.> Beanstalkd is a task queue which supports three modes of message
> delivery:


>  * At-least once delivery: The most common one. A failed message/task
>    will get redelivered and you make sure to have idempotent tasks
>    that can be retried gracefully.
>  * At-most once delivery: If you delete a message/task immediately
>    after reserving it.
>  * Exactly once delivery: You store the job ID in downstream databases
>    you are modifying. That way you'll know if you have processed the
>    job previously. You'll likely want to use the write-ahead log for
>    this for job ids to remain the same on restart.> The above examples are assuming you have not published duplicates of a
> task (which is what this PR tries to solve). To make sure you don't
> submit duplicates, I propose you generate a unique ID in your task
> body and use that in your task workers to deduplicate your work (using
> similar strategy as described in bullet list above). Would that work?
> If so, let me know a case it doesn't solve and I'll reopen this PR.> 
> In job.c[1]:


> > @@ -89,6 +89,27 @@ job_find(uint64 job_id) return jh; }  +job
> > +job_find_by_body(job j)
>> This function is O(n) where n is number of jobs in queue, right?
> Unfortunately, that's not acceptable. We want at least O(log(n))
> or better.> — You are receiving this because you authored the thread. Reply to
> this email directly, view it on GitHub[2], or mute the thread[3].

Links:

  1. https://github.com/beanstalkd/beanstalkd/pull/180#discussion_r211104749
  2. https://github.com/beanstalkd/beanstalkd/pull/180#pullrequestreview-147460651

Issue #112, opened 2012-04-15
Support sunos/illumos


jbergstroem:
Consider this a tracking bug in adding support for the illumos (sunos is secondary target) platform

PR #202, opened 2013-08-12
add sunos.c for solaris/illumos support

potatosalad:
I added preliminary solaris/illumos support with event ports in `sunos.c` as mentioned in #112.  The other changes outside `sunos.c` were related to portability issues between BSD/Linux and SunOS.

This branch was tested with `make bench` on the following platforms:
- SmartOS joyent_20130629T040542Z with GCC 4.7.2
- CentOS+OpenVZ 2.6.32-042stab076.7 with GCC 4.7.1
- OS X 10.8.4 with clang 4.2

Here's the related documentation for event ports:
- [port_create(3c)](http://illumos.org/man/3c/port_create)
- [port_associate(3c)](http://illumos.org/man/3c/port_associate)
- [port_getn(3c)](http://illumos.org/man/3c/port_getn)
- [poll(2)](http://illumos.org/man/2/poll)

### DIFF
--- a/conn.c
+++ b/conn.c
@@ -37,13 +37,13 @@ make_conn(int fd, char start_state, tube use, tube watch)
     job j;
     Conn *c;
 
-    c = new(Conn);
-    if (!c) return twarn("OOM"), NULL;
+    c = (Conn *)new(Conn);
+    if (!c) return twarn("OOM"), (Conn *)NULL;
 
     ms_init(&c->watch, (ms_event_fn) on_watch, (ms_event_fn) on_ignore);
     if (!ms_append(&c->watch, watch)) {
         free(c);
-        return twarn("OOM"), NULL;
+        return twarn("OOM"), (Conn *)NULL;
     }
 
     TUBE_ASSIGN(c->use, use);
--- a/conn.c
+++ b/conn.c
@@ -37,13 +37,13 @@ make_conn(int fd, char start_state, tube use, tube watch)
     job j;
     Conn *c;
 
-    c = (Conn *)new(Conn);
-    if (!c) return twarn("OOM"), (Conn *)NULL;
+    c = new(Conn);
+    if (!c) return twarn("OOM"), NULL;
 
     ms_init(&c->watch, (ms_event_fn) on_watch, (ms_event_fn) on_ignore);
     if (!ms_append(&c->watch, watch)) {
         free(c);
-        return twarn("OOM"), (Conn *)NULL;
+        return twarn("OOM"), NULL;
     }
 
     TUBE_ASSIGN(c->use, use);
--- a/testserv.c
+++ b/testserv.c
@@ -59,7 +59,7 @@ muststart(char *a0, char *a1, char *a2, char *a3, char *a4)
 
     /* now in child */
 
-    execlp(a0, a0, a1, a2, a3, a4, NULL);
+    execlp(a0, a0, a1, a2, a3, a4, (char*)NULL);
 }
 
 static int
--- a/testserv.c
+++ b/testserv.c
@@ -59,7 +59,7 @@ muststart(char *a0, char *a1, char *a2, char *a3, char *a4)
 
     /* now in child */
 
-    execlp(a0, a0, a1, a2, a3, a4, (char*)NULL);
+    execlp(a0, a0, a1, a2, a3, a4, NULL);
 }
 
 static int
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,10 @@ LDLIBS?=
 OS?=$(shell uname | tr 'A-Z' 'a-z')
 INSTALL?=install
 
+ifeq ($(OS),sunos)
+LDFLAGS += -lxnet -lsocket -lnsl
+endif
+
 VERS=$(shell ./vers.sh)
 TARG=beanstalkd
 MOFILE=main.o
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ OS?=$(shell uname | tr 'A-Z' 'a-z')
 INSTALL?=install
 
 ifeq ($(OS),sunos)
-LDFLAGS += -lxnet -lsocket -lnsl
+override LDFLAGS += -lxnet -lsocket -lnsl
 endif
 
 VERS=$(shell ./vers.sh)
--- a/dat.h
+++ b/dat.h
@@ -38,6 +38,16 @@ typedef int(FAlloc)(int, int);
 #define NUM_PRIMES 19
 #endif
 
+/* Some compilers (e.g. gcc on SmartOS) define NULL as 0.
+ * This is allowed by the C standard, but is unhelpful when
+ * using NULL in most pointer contexts with errors turned on. */
+#if (defined(sun) || defined(__sun)) && (defined(__SVR4) || defined(__svr4__))
+#ifdef NULL
+#undef NULL
+#endif
+#define NULL ((void*)0)
+#endif
+
 // The name of a tube cannot be longer than MAX_TUBE_NAME_LEN-1
 #define MAX_TUBE_NAME_LEN 201
 
--- a/sunos.c
+++ b/sunos.c
@@ -1,0 +1,111 @@
+#include <stdint.h>
+#include <fcntl.h>
+#include <stdlib.h>
+#include <sys/types.h>
+#include <unistd.h>
+#include <errno.h>
+#include <poll.h>
+#include <port.h>
+#include "dat.h"
+
+static int portfd;
+static char buf0[512]; /* buffer of zeros */
+
+/* Allocate disk space.
+ * Expects fd's offset to be 0; may also reset fd's offset to 0.
+ * Returns 0 on success, and a positive errno otherwise. */
+int
+rawfalloc(int fd, int len)
+{
+    int i, w;
+
+    for (i = 0; i < len; i += w) {
+        w = write(fd, buf0, sizeof buf0);
+        if (w == -1) return errno;
+    }
+
+    lseek(fd, 0, 0); /* do not care if this fails */
+
+    return 0;
+}
+
+
+int
+sockinit(void)
+{
+    portfd = port_create();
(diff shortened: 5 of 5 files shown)


kr:
This looks good, but I am worried about ongoing maintenance, since I don't have ready access to Solaris/Illumos/SmartOS systems to test. I am happy to merge it anyway, but it'll need some folks to keep an eye on it if changes ever need to happen to the Socket interface.

Do you mind rebasing this? A few of the miscellaneous fixes here have made their way in already.

JensRantil:
In the best of worlds someone could contribute a machine running build+unit tests on Solaris/Illumos/SmartOS so we could get feedback in the pull requests.

potatosalad:
I've rebased my changes onto master and checked that the build and unit tests still work correctly on a Joyent public cloud instance:
packager 0d7d688b-50f4-406f-ab94-7654bb3af000 1 2016-04-03 09-27-50
There's probably some way of automating the builds using an ephemeral server on Joyent for each pull request, but I'm not sure whether there's an existing service for that out there or if one would need to be built.

Alternatively, changing all of the socket related operations to the abstractions provided by a third-party event library like [libevent](http://libevent.org/) or [libuv](http://libuv.org/) might reduce the complexities of having to manually deal with supporting each of the different event systems at the expense of adding an external dependency.

kr:
> Alternatively, changing all of the socket related operations to the abstractions provided by a third-party event library like libevent or libuv

Beanstalkd originally used libevent. It was more trouble than it was worth.

I'm not *that* worried about ongoing maintenance. The other platforms have been pretty hassle-free once the initial bugs are shaken out. Presumably Solaris isn't so bizarre that the same won't hold there as well.

conn.c
```diff
@@ -33,13 +33,13 @@ make_conn(int fd, char start_state, tube use, tube watch)
     job j;
     Conn *c;
 
-    c = new(Conn);
-    if (!c) return twarn("OOM"), NULL;
+    c = (Conn *)new(Conn);
```
kr: The changes in this file seem unnecessary. The return value of macro `new` is `void*`, which doesn't require a cast to convert to another type. Same goes for testserv.c.

If it does turn out to be necessary, let's do it in a separate pull request.

potatosalad: This is the error returned without these changes:
beanstalkd beanstalkd-build 1 2016-04-03 21-27-43
This is with GCC 4.7.4 on SmartOS joyent_20160317T000621Z.

testserv.c
```diff
@@ -55,7 +55,7 @@ muststart(char *a0, char *a1, char *a2, char *a3, char *a4)
 
     /* now in child */
 
-    execlp(a0, a0, a1, a2, a3, a4, NULL);
+    execlp(a0, a0, a1, a2, a3, a4, (char*)NULL);
```
potatosalad: The error thrown here without this change is:
beanstalkd beanstalkd-build 1 2016-04-03 21-31-38
This is with GCC 4.7.4 on SmartOS joyent_20160317T000621Z.

conn.c
kr: Out of curiosity, what happens if you run

```
cc -DNULL='((void*)0)' -Wall -Werror -Wformat=2 -g    -c -o conn.o conn.c
```

on the original file without these changes?

potatosalad: It compiles without an error.  Same thing for `testserv.c`.  Would adding `-DNULL='((void*)0)'` to the `CFLAGS` when sunos is detected be a better solution?

kr: I think the thing to do is put something in dat.h like this, with a comment

```
// work around unhelpful built-in NULL on some compilers
#ifndef NULL
#define NULL ((void*)0)
#endif
```

If that still gets the job done cleanly on your system, let's roll with it.

potatosalad:
I tried the `#ifndef NULL`, but I guess `NULL` is actually defined as something because I was unable to get it to work.  Instead, I added this to `dat.h` which works as expected:

```c
#if (defined(sun) || defined(__sun)) && (defined(__SVR4) || defined(__svr4__))
#ifdef NULL
#undef NULL
#endif
#define NULL ((void*)0)
#endif
```

The long `#if` statement is to distinguish between Solaris and SunOS (which are different operating systems, despite Solaris having the `uname` of `SunOS` with `SVR4` features added...super confusing).

potatosalad:
Also, I just realized my commit message is the exact opposite of what was actually done.

potatosalad:
Is there anything else that needs to be fixed or changed for this PR to be accepted?

ysmolsky:
@potatosalad are you still interested in pushing this PR?

potatosalad:
Sure, from first glance I don't see any actual code changes that would need to take place beyond rebasing things from `master`.

ysmolsky:
LGTM. But I have one question. How can I automate testing of beanstalkd on SunOS? Is there a way? I am really concerned that I must a change that will break stuff on SunOS and I will have no idea about it. I am that careful.

ysmolsky:
One last request, could you please run `make clean bench` on SunOS and paste the output here? I would like to see how it performs there.

potatosalad:
> LGTM. But I have one question. How can I automate testing of beanstalkd on SunOS? Is there a way? I am really concerned that I must a change that will break stuff on SunOS and I will have no idea about it. I am that careful.

Outside of running something like Jenkins, I'm not sure if there's a good way to automate testing against SunOS/Solaris/illumos.

> One last request, could you please run `make clean bench` on SunOS and paste the output here? I would like to see how it performs there.

Using [VMware Fusion](https://www.vmware.com/products/fusion.html) and [SmartOS](https://wiki.smartos.org/download-smartos/) version `joyent_20190718T005708Z` the results are:

```
# make clean bench
rm -f vers.c  beanstalkd job.o tube.o walg.o testjobs.o time.o testutil.o heap.o prot.o testserv.o util.o sd-daemon.o serv.o vers.o sunos.o ms.o conn.o testheap.o primes.o file.o net.o ct/_ctcheck.c ct/_ctcheck.o ct/ct.o ct/_ctcheck.o
cc -Wall -Werror -Wformat=2 -g   -c -o testheap.o testheap.c
cc -Wall -Werror -Wformat=2 -g   -c -o testjobs.o testjobs.c
cc -Wall -Werror -Wformat=2 -g   -c -o testserv.o testserv.c
cc -Wall -Werror -Wformat=2 -g   -c -o testutil.o testutil.c
ct/gen testheap.o testjobs.o testserv.o testutil.o  >ct/_ctcheck.c.part
mv ct/_ctcheck.c.part ct/_ctcheck.c
cc -Wall -Werror -Wformat=2 -g   -c -o ct/_ctcheck.o ct/_ctcheck.c
cc -Wall -Werror -Wformat=2 -g   -c -o ct/ct.o ct/ct.c
cc -Wall -Werror -Wformat=2 -g   -c -o sunos.o sunos.c
cc -Wall -Werror -Wformat=2 -g   -c -o conn.o conn.c
cc -Wall -Werror -Wformat=2 -g   -c -o file.o file.c
cc -Wall -Werror -Wformat=2 -g   -c -o heap.o heap.c
cc -Wall -Werror -Wformat=2 -g   -c -o job.o job.c
cc -Wall -Werror -Wformat=2 -g   -c -o ms.o ms.c
cc -Wall -Werror -Wformat=2 -g   -c -o net.o net.c
cc -Wall -Werror -Wformat=2 -g   -c -o primes.o primes.c
cc -Wall -Werror -Wformat=2 -g   -c -o prot.o prot.c
cc -Wall -Werror -Wformat=2 -g   -c -o sd-daemon.o sd-daemon.c
cc -Wall -Werror -Wformat=2 -g   -c -o serv.o serv.c
cc -Wall -Werror -Wformat=2 -g   -c -o time.o time.c
cc -Wall -Werror -Wformat=2 -g   -c -o tube.o tube.c
cc -Wall -Werror -Wformat=2 -g   -c -o util.o util.c
./verc.sh >vers.c
cc -Wall -Werror -Wformat=2 -g   -c -o vers.o vers.c
cc -Wall -Werror -Wformat=2 -g   -c -o walg.o walg.c
cc  -lxnet -lsocket -lnsl  ct/_ctcheck.o ct/ct.o sunos.o conn.o file.o heap.o job.o ms.o net.o primes.o prot.o sd-daemon.o serv.o time.o tube.o util.o vers.o walg.o testheap.o testjobs.o testserv.o testutil.o   -o ct/_ctcheck
ct/_ctcheck -b
.........................................................................

PASS
ctbench_heap_insert	 3000000	      1109 ns/op
ctbench_heap_remove	 2000000	      1002 ns/op
ctbench_make_job	 5000000	      1202 ns/op
ctbench_put_delete_1k	   10000	    208513 ns/op	   5.10 MB/s
ctbench_put_delete_64k	    3000	    612422 ns/op	 181.42 MB/s
ctbench_put_delete_8	   10000	    190305 ns/op	   0.07 MB/s
ctbench_put_delete_8k	   10000	    249182 ns/op	  39.98 MB/s
```

Same machine with the non-virtualized host macOS 10.14.5:

```
# make clean bench
.........................................................................

PASS
ctbench_heap_insert	 5000000	       439 ns/op
ctbench_heap_remove	 2000000	       837 ns/op
ctbench_make_job	10000000	       210 ns/op
ctbench_put_delete_1k	   10000	    208889 ns/op	   5.10 MB/s
ctbench_put_delete_64k	    3000	    391736 ns/op	 193.22 MB/s
ctbench_put_delete_8	    5000	    231966 ns/op	   0.04 MB/s
ctbench_put_delete_8k	    5000	    247555 ns/op	  40.01 MB/s
```

potatosalad:
@ysmolsky On the CI/CD side of things, do you have a technique you're using to check FreeBSD?  That same technique might be reusable to test against SunOS.

ysmolsky:
This project is not tested against FreeBSD yet. I was looking at options besides setting up it manually and paying for the instance.

ysmolsky:
Thank you very much for benchmarks. This prove that on linux performance of beanstalkd is impaired contrary to every other platform.

PR #203, opened 2013-08-28
Add Command: "wait <id>" that responds when a job is deleted

danielbeardsley:
Add a command: `wait <id>\r\n` that responds with `DELETED <id>\r\n` or `BURIED <id>\r\n` only once a job has been deleted or buried.

This allows a client to be notified about the status of a job without polling.
### From the docs

The wait command waits until a particular job is out of the system before responding. It can be used to block a client until a job is completed. It looks like this:

```
wait <id>\r\n
```
##### `wait` options
- `<id>` is the job id to wait for.
##### `wait` responses

The client expects one line of response once the state of the job has changed, which may be:
- `DELETED <id>\r\n` to indicate the job was deleted via the `delete` command (normal success).
- `BURIED <id>\r\n` to indicate the job was buried via the `bury` command.
- `NOT_FOUND\r\n` if the job id does not exist.

A response won't be sent by the server until the specified job is deleted or buried. The wait command can be issued multiple times before a response is received in order to wait on a collection of jobs.

### DIFF
--- a/conn.c
+++ b/conn.c
@@ -51,6 +51,8 @@ make_conn(int fd, char start_state, tube use, tube watch)
     c->tickpos = -1;
     j = &c->reserved_jobs;
     j->prev = j->next = j;
+    c->next_waitjob = NULL;
+
 
     /* stats */
     cur_conn_ct++;
--- a/dat.h
+++ b/dat.h
@@ -149,6 +149,7 @@ struct job {
     void *reserver;
     int walresv;
     int walused;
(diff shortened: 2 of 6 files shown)


ysmolsky:
So many things have changed since this PR was open. Also I have some doubts regarding the design and interface of this command. I do not think we should progress with this PR. I have opened issue #560 that records this proposal and links to this PR in case of somebody wants to resurrect it.

PR #207, opened 2013-10-02
Unit Tests: get ct compiling on CentOS-5

danielbeardsley:
Fix some bugs that were preventing compilation (at least on CentOD 5).

Add `-lrt` to load the rt libary to prevent this error:
> ct.c:(.text+0x12): undefined reference to `clock_gettime'

Include `time.h` to prevent this error:
> ct/ct.c:47: error: ‘CLOCK_MONOTONIC’ undeclared (first use in this function)

Perform a cast to prevent this error: (on a 64bit arch)
> ct/ct.c:420: warning: format ‘%10lld’ expects type ‘long long int’, but argument 3 has type ‘int64

Don't use an expression in a const definition to prevent this error:
> ct/ct.c:28: error: initializer element is not constant

See pull kr/ct#7

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,7 @@ check: ct/_ctcheck
 bench: ct/_ctcheck
 	ct/_ctcheck -b
 
-ct/_ctcheck: ct/_ctcheck.o ct/ct.o $(OFILES) $(TOFILES)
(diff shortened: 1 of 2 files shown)


kr:
Updating ct just happened in 8a5665ae9a6594da68b9519b261c0328455bb1a5.

Adding librt went in 154162201fe498b9d3b83869199ac33cb5ab7dfa.

Sorry for not getting to this PR sooner.

PR #216, opened 2013-11-11
Increase LINE_BUF_SIZE to handle longest possible command

tdg5:
This fix addresses https://github.com/kr/beanstalkd/issues/211 and https://github.com/kr/beanstalkd/issues/212. 

I tried the approach I suggested in the issue of changing prot.c:343 to use ```>``` instead of ```>=```, but trying to use a tube with a name of length 200 continued having issues. Also given the number of other commands having length issues, this approach would not have solved the larger problem.

So this instead updates LINE_BUF_SIZE to handle the longest possible command and reply lines.

This may be a preferable approach since both LINE_BUF_SIZE and MAX_TUBE_NAME_LEN follow similar logic in that they both represent the minimum invalid length.

Existing ct tests all pass. Additionally a suite of ~200 integration tests I wrote to test adherence to the beanstalk protocol show no regressions. Those tests can be found here: https://github.com/tdg5/beanstalk_integration_tests. They're written in ruby though, sorry :)

### DIFF
--- a/dat.h
+++ b/dat.h
@@ -43,9 +43,9 @@ typedef int(FAlloc)(int, int);
 #define MAX_TUBE_NAME_LEN 201
 
 /* A command can be at most LINE_BUF_SIZE chars, including "\r\n". This value
- * MUST be enough to hold the longest possible command or reply line, which is
- * currently "USING a{200}\r\n". */
-#define LINE_BUF_SIZE 208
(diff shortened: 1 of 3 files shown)


kr:
This looks great. Thanks for the fix! Do you mind rebasing this? It doesn't merge cleanly any more.

tdg5:
Rebased!

PR #217, opened 2013-11-12
pause-tube should check tube name is ok

tdg5:
This change updates pause-tube to return BAD_FORMAT when an invalid tube name is supplied. At present pause-tube will return NOT_FOUND after it fails to find an invalid tube.

This is more of a consistency thing than a bug, per se. 

ct tests all pass.


kr:
thanks for the fix!

PR #218, opened 2013-11-13
limit the update frequency of the epoll/kqueue timeout period, fixes #200

### LINKED ISSUE #200: Beanstalk 1.9 high CPU usage
I've decided to open a new issue with the beanstalk v1.9 performance issues detected by several users.

https://groups.google.com/forum/#!searchin/beanstalk-talk/performance/beanstalk-talk/DSDOE7spoio/a8l3jO0wjG0J

I've ran a callgrind benchmark on version v1.9 and v1.8, and from what I can see that due to the changes from 1d191ba26b20f402cc8e8ee3a7f7b0de9f1ff78c, prottick is called more often from srvserve which seems to be causing a higher CPU usage.
I hope this helps to debug the problems.
profiling_1_9vs1_8

To run the profile I've used this test program https://github.com/swehner/bsperf similar to the one matt used when sending this issue to the mailing list.

nephics:


Roguelazer:
I don't think this patch is quite correct; it makes the new unit tests in 1.10 fail, among other things.

I think the right fix is to force `prottick` to run if there was any socket activity by setting `t = 0` in the `if (rw)` block. Patch to your patch is:

```
diff --git a/serv.c b/serv.c
index 5eb2d10..c63cee5 100644
--- a/serv.c
+++ b/serv.c
@@ -61,6 +61,7 @@ srvserve(Server *s)

         if (rw) {
             sock->f(sock->x, rw);
+            t = 0;
         }
     }
 }
```

Roguelazer:
Bump?

ysmolsky:
The author of this PR has deleted his account. @Roguelazer , would you be so kind to take over this and look into it? Also I am interested if this is reproducible on master. If so then we should abandon this PR and create a new one that fixes the issue.

ysmolsky:
Closing since author has deleted his account.

Issue #145, opened 2012-10-11
how should we know if the beastalkd service is in drain mode?


JerryKwan:
is there any way we can tell the beanstalkd service is  in drain mode?
when we try to put a new job, may be we can tell the beanstalkd service is in drain mode through the return result. if we can add some info in stats command?then we can tell the drain mode at any time without try to input some job?

chrisminett:
There's a commit for this that appears to be orphaned. Any chance of seeing this put into a release?

JensRantil:
Typo in issue title.

ysmolsky:
> There's a commit for this that appears to be orphaned. Any chance of seeing this put into a release?

@chrisminett what is the commit?

chrisminett:
@ysmolsky no idea, that was 4 years ago!

At a guess it was probably something to do with #219 which has now been replaced by #307 , to which you're assigned. The change to the `stats` command achieved by that PR looks good for our use case - being able to tell if a server is draining without having to put a job.

ysmolsky:
@chrisminett thanks, this helps me. I will look into that PR and come with my own, better, fix, hopefully.

PR #219, opened 2013-12-06
add draining_mode status to the `stats` command #145

justincase:
```
binlog-max-size: 10485760
id: 42cdbb04ec381e89
hostname: Mac.local
status: OK
```

`kill -SIGUSR1 <pid>`

```
binlog-max-size: 10485760
id: 42cdbb04ec381e89
hostname: Mac.local
status: DRAINING
```

All tests passes.

### DIFF
--- a/dat.h
+++ b/dat.h
@@ -372,11 +372,15 @@ int  filewrjobfull(File*, job);
 
 
 #define Portdef "11300"
+#define Addrdef NULL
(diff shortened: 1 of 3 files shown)


kr:
Thanks for doing this! The feature looks useful, but I think it would be more straightforward to put `draining: false` or `draining: true` in the output. This should be a 2-line change.

PR #249, opened 2014-10-09
Subtree ct

decimalbell:
include subproject ct using git subtree


kr:
I think the right way to do this is to take just the ct subdirectory from the upstream repo and subtree that, rather than the root.

ysmolsky:
It's much easier to copy this repo once in a while rather than fiddle with submodules or subtrees. They are clunky and somewhat annoying. I am all ears to hear arguments in the form of good looking patch.  Closing down this patch for now.

PR #254, opened 2014-11-08
Properly catch and report getaddrinfo() errors

### LINKED ISSUE #253: Segmentation fault when using an invalid listen address
When launching _beanstalkd_ with an incorrect address for option `-l`, the process terminates abruptly with a segmentation fault.

etanol:
The getaddrinfo() function does not behave like a system call.  In particular,
the only successful return value is zero; any other value should be considered a
failure, not only -1.

This commit fixes a segfault caused by an invalid value of the "-l" command line
option.

This fixes #253


silviucpp:
@etanol  can you please update this patch to handle the current master?

net.c
```diff
@@ -48,8 +48,10 @@ make_server_socket(char *host, char *port)
     hints.ai_socktype = SOCK_STREAM;
     hints.ai_flags = AI_PASSIVE;
     r = getaddrinfo(host, port, &hints, &airoot);
-    if (r == -1)
-      return twarn("getaddrinfo()"), -1;
+    if (r != 0) {
```
pprindeville: Much better, and using comma notation is an abomination.

PR #275, opened 2015-08-21
add protocol chinese version

mickeyouyou:
as title , we use beanstlkd in our project ,so we translate it into chinese for more chinese developer.

I have sent a email to @kr before.


kr:
Merged! Thank you so much. :100: :8ball: :balloon:

mickeyouyou:
:smiley: 

:thumbsup:

PR #278, opened 2015-10-04
Use 64 bit ints for bytes used/reserved and ratio

Minnozz:
This fixes an integer overflow in the free/used ratio calculation for
binlog compaction, and allows for tracking more than ~2.1GB of free or
used data in beanstalkd.

Fixes kr/beanstalkd#276

### DIFF
--- a/walg.c
+++ b/walg.c
@@ -94,7 +94,7 @@ ratio(Wal *w)
     int64 n, d;
 
     d = w->alive + w->resv;
-    n = w->nfile*w->filesize - d;
+    n = (int64)w->nfile * (int64)w->filesize - d;
     if (!d) return 0;
     return n / d;
 }
--- a/walg.c
+++ b/walg.c
@@ -91,7 +91,7 @@ usenext(Wal *w)
 static int
 ratio(Wal *w)
 {
-    int n, d;
+    int64 n, d;
 
     d = w->alive + w->resv;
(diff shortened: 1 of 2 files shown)


JensRantil:
Somewhat related from someone who don't know the code specifically; What will the new limit be? Is there any way to warn _before_  any of these limits are hit? Should we?

kr:
> Somewhat related from someone who don't know the code specifically; What will the new limit be? Is there any way to warn before any of these limits are hit? Should we?

Using a 64-bit int means there will be no practical limit.

walg.c
```diff
@@ -91,7 +91,7 @@ usenext(Wal *w)
 static int
 ratio(Wal *w)
 {
-    int n, d;
+    int64 n, d;
 
     d = w->alive + w->resv;
     n = w->nfile*w->filesize - d;
```
kr: I think filesize will also have to be 64 bits, and this expression should convert nfile to int64 explicitly, to make sure the multiply is done in 64 bits.

Minnozz:
I added the cast to 64 bits, but changing filesize to 64 bits too involves more changes that I have not had time for yet (for instance the command line argument parsing).

kr:
Cool, that works. Thanks.

kr:
Thanks so much for the fix!

PR #279, opened 2015-10-05
Code Restructure

handautsav:
#### What's this PR do?
It structures the entire project into standard directory structure including source code, tests, documentation, extras etc.

#### Where should the reviewer start?
I would recommend to start with initial 'src' directory structuring and move up to 'tests' structure and integration.

#### Any background context you want to provide?
I have recently started using beanstalkd in a personal project and found the source code could use some help in restructuring and take it forward.

#### Current Structure
- src                     source code
- src/ct                  testing tool; see https://github.com/kr/ct
- doc                     documentation
- pkg                     miscellaneous files for packaging
- extras/init-scripts     files useful for system administrators


kr:
Thanks for your interest! I'm glad you found beanstalkd worthwhile and interesting enough to contribute to. It's generally a good idea to discuss proposed changes, especially refactorings, first to avoid spending too much time on something that's unlikely to get merged. The mailing list is the best place for such discussions. (Github pull requests aren't a very good discussion forum.)

Do you have specific problems you're trying to solve? Right now, this patch appears to be superficial, just moving things around without a clear purpose. It would add clutter to the project's history without substantial benefit, so I can't merge this as it is. However, there might be another way to solve your problem. There's nothing wrong with refactoring and renaming things if the change is focused and well justified. If you want to articulate what your goals are, we can work together to find a good solution.

handautsav:
I strongly believe in code organization for clarity and structure. The source code layout should reflect the architecture; as a corollary, an organized code-base is well-structured source layout. The proposed structure 'groups' the source code, but the idea is to be able to quickly identify layout. For example, source code had been moved into "src" folder, Tests had been moved to "src/tests" folder. This is also particularly useful and easier to explore than if all the files sit in the same folder.

I would be happy to discuss this in here or project mailing list.

kr:
Okay, if I understand correctly, the goal here really is to make the overall file layout of the project easier to understand and navigate.

I agree that some variation of this change would make the root directory of the project look tidier, but I have a hard time believing the source code would be that much easier to find if it were moved into a subdirectory. :) There are also some minor drawbacks, such as the extra keystrokes it would take to navigate. I was about to say something about what the net effect would be "on balance", but it probably makes not enough difference in either direction to be measurable.

Issues such as this are, unfortunately, mostly a matter of taste, and therefore the ideal varies from person to person. This is a topic ripe for [bike shed](https://en.wikipedia.org/wiki/Parkinson%27s_law_of_triviality) discussions, and the most humane thing to do in such a situation is to end the discussion quickly. So I will, regretfully, decline this patch.

But again, thank you! And I hope this hasn't been a discouraging experience. If you want to contribute to more open source projects in the future, please do!

PR #280, opened 2015-10-06
Makefile: resolved parallel install issue

dafugg:
When installing within [Open Embedded](http://www.openembedded.org/) the make jobs parameter is by default shared between "all" and "install" [as per this doc](http://www.yoctoproject.org/docs/latest/mega-manual/mega-manual.html#var-PARALLEL_MAKEINST). The beanstalkd Makefile was causing needless parallelization where the beanstalkd binary may be installed before the directory is created.


kr:
:+1:

PR #208, opened 2013-10-09
Document stats-job values delay and ttr

tdg5:
Document delay and ttr values returned by stats-job command. Language borrowed from elsewhere in the docs.

PR #285, opened 2015-11-12
Update protocol.txt - stats-job & ttr

Perlover:
From protocol.txt:
"Both the TTR and the actual time left can be
found in response to the stats-job command."
I saw perl module - there is too in stats-job should be "ttr" stat info
But in the protocol.txt for stats-job there is not "ttr" description. I think it's doc's error


Perlover:
After creation of this pull request i found dupliacate pull request:

https://github.com/kr/beanstalkd/pull/208

kr:
Thanks! I merged #208 because it also included a description of `delay`.

PR #289, opened 2016-01-04
Flag to put beanstalkd into the background

ptjm:
This change adds a flag (-D), which causes beanstalkd to fork off a child process and print its PID to stdout from the parent process. The difference between this and running beanstalkd in the background is the process forked with the -D flag calls setsid(), which dissociates it from the controlling terminal and starts a new process group.

I couldn't think of a way to unit test this change.

### DIFF
--- a/dat.h
+++ b/dat.h
@@ -378,6 +378,8 @@ struct Server {
     char *addr;
     char *user;
 
+    int   bg;
+
     Wal    wal;
     Socket sock;
     Heap   conns;
--- a/doc/beanstalkd.ronn
+++ b/doc/beanstalkd.ronn
@@ -35,6 +35,10 @@ and format of the `beanstalkd` protocol.
   (Do not use this option, except to negate `-n`. Both `-c` and `-n`
   will likely be removed in a future `beanstalkd` release.)
 
+* `-D`:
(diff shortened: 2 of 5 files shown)


kr:
Beanstalkd used to have this feature, but it was removed in 2e24666ccc009022ec721ce019bd2461c4dbcb8d. In principle, this sort of thing is better done by a separate tool or the operating system; it is unrelated to beanstalkd's purpose. In practice, there are other tools out there that do a much more thorough job of supervising processes, for example, dbj's [daemontools](https://cr.yp.to/daemontools.html), and [runit](http://smarden.org/runit/), as well as services built in to operating systems, such as [launchd](https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man8/launchd.8.html) and [systemd](https://wiki.freedesktop.org/www/Software/systemd/).

If all you want is to call setsid, on a Linux system consider using the command [setsid (1)](http://man.cx/setsid(1)).

kr:
I should also say thank you for doing the work to send this change! It is really appreciated. I'm sorry that the design philosophy of beanstalkd precludes this patch, but please do consider picking through the open issues to find something that can be improved.

PR #290, opened 2016-01-08
include time.h for clock_gettime function definitions

gnat42:
I've taken package maintainership in fedora / epel. One of many compilation error messages is that clock_gettime is implicitly defined. Its fixed by including <time.h> as this patch does.


kr:
Thanks! This has actually been fixed in the upstream source of ct. The best way to fix this is to update to the newest version. Done in 8a5665ae9a6594da68b9519b261c0328455bb1a5.

PR #291, opened 2016-01-08
static const compilation error

gnat42:
Fedora has strict compilation flags. The static const int64 isn't strictly valid C apparently. With this change the compilation succeeds in fedora.


kr:
Thanks! This has actually been fixed in the upstream source of ct. The best way to fix this is to update to the newest version. Done in 8a5665ae9a6594da68b9519b261c0328455bb1a5.

PR #292, opened 2016-01-08
use more secure mkdtemp

gnat42:
during compilation, a message stating that mkdtemp is more secure than mktemp. Here's a patch to make that change.


kr:
Thanks! This has actually been fixed in the upstream source of ct. The best way to fix this is to update to the newest version.

PR #293, opened 2016-01-08
printing int64 on different archs

gnat42:
There are warnings about size mis-matches. Including inttypes.h and using the constant fixes the issue so that the print format is the same regardless of x86_64 or i686.

### DIFF
--- a/ct/ct.c
+++ b/ct/ct.c
@@ -14,6 +14,7 @@
 #include <errno.h>
 #include <sys/time.h>
 #include <stdint.h>
+#include <inttypes.h>
 #include "internal.h"
 #include "ct.h"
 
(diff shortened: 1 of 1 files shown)


kr:
Thanks! This has actually been fixed in the upstream source of ct. The best way to fix this is to update to the newest version.

sergeyklay:
@kr How about release new version?

JensRantil:
> How about release new version?

@sergeyklay Please track #383.

ct/ct.c
```diff
@@ -14,6 +14,7 @@
 #include <errno.h>
 #include <sys/time.h>
 #include <stdint.h>
+#include <inttypes.h>
```
pprindeville: Minor comment.  I'd add something like:

```
#ifndef PRId64
#error "Need definition for PRId64 but not provided by <stdint.h>"
#endif
```

pprindeville: Oops, missed that a similar fix was merged as part of a different commit...

PR #295, opened 2016-01-28
Reduce the size of the hashmap when it's too sparse

Shir0kamii:
Hi,

In my team, we use beanstalkd a lot and we found that it holds a large quantity of memory if it has been filled with a large collections once, even if it's empty afterwards.

I investigated the code and found out that beanstalkd keep a hash map relating ids to their job. This hash map is extended when at least a quarter of it is filled, but it never tries to shrink. So I implemented this shrinking, with a factor of 1/16 to prevent the hash map from scaling too often.

I hope it helps !


kr:
This is great! Thanks for tracking this down and doing the work to fix it. There are just a couple of small things we need to fix and then get this merged. :grin: Inline comments to follow.

Makefile
```diff
@@ -64,6 +64,10 @@ $(OFILES) $(MOFILE): $(HFILES)
 clean:
 	rm -f *.o $(CLEANFILES)
 
+.PHONY: fclean
+fclean: clean
```
kr: Let's undo the changes to this file. If you want to tweak the build rules, it can happen in a separate pull request.

README
```diff
@@ -1,5 +1,11 @@
-This is beanstalkd, a fast, general-purpose work queue.
-See http://kr.github.io/beanstalkd/ for general info.
+This is a fork of kr/beanstalkd, a fast, general-purpose work queue focusing
+on features and fixes addressing performance and production critical problems.
+
+
+ADDED FEATURES AND FIXES
```
kr: Do you mind keeping these changes out of the PR; I can't merge them. Of course, it still makes sense to update the readme on your fork, but I guess it should be on another branch.

job.c
```diff
@@ -34,7 +34,7 @@ store_job(job j)
     all_jobs_used++;
 
     /* accept a load factor of 4 */
-    if (all_jobs_used > (all_jobs_cap << 2)) rehash();
+    if (all_jobs_used > (all_jobs_cap >> 1)) rehash();
```
kr: Let's put this back how it was. The intention is to accept a [load factor](https://en.wikipedia.org/wiki/Load_factor_(computer_science)) of 4, so we grow the hash table when the number of entries exceeds the 4*capacity.

job.c
```diff
@@ -135,6 +169,9 @@ job_hash_free(job j)
         *slot = (*slot)->ht_next;
         --all_jobs_used;
     }
+
+    // Downscale when the hashmap is too sparse
+    if (all_jobs_used < (all_jobs_cap >> 3) && cur_prime > 0) rehash_down();
```
kr: The cur_prime > 0 check here is redundant with the one inside rehash_down. Let's remove this one.

job.c
```diff
@@ -44,13 +44,13 @@ rehash()
...
     if (cur_prime >= NUM_PRIMES) return;
-    if (hash_table_was_oom) return;
+    if (hash_table_max_prime <= cur_prime) return;
 
     all_jobs_cap = primes[++cur_prime];
     all_jobs = calloc(all_jobs_cap, sizeof(job));
     if (!all_jobs) {
         twarnx("Failed to allocate %zu new hash buckets", all_jobs_cap);
-        hash_table_was_oom = 1;
+        hash_table_max_prime = cur_prime;
```
kr: This won't work. A calloc failure here doesn't mean we've reached a maximum size for the hash table, it means the process *doesn't have enough memory* to allocate the new table. It's because of all the other objects (jobs, buffers, whatever) that are allocated at the moment, using up all that memory.

We could just clean up and return here, and then the next job added would immediately try to grow the hash table again, and likely fail, and same for the job after that, constantly trying (and failing) to allocate a new hash table, until a bunch of memory gets freed and then maybe it works again, or else jobs have been deleted so we don't actually need to grow the hash table any more.

That would be correct, but I had thought the call to calloc would be expensive, even if it fails, so as a simple hack to prevent that scenario of constantly failing to grow the hash table, I made the failure sticky. If it ever fails, it just gives up permanently and accepts whatever load factor fate bestows upon us.

That simple hack might not be good enough now that we're also trying to *shrink* the hash table. Presumably while deleting jobs we're also freeing memory, so if calloc fails in rehash_down, it might succeed the next time, or the time after that. I'd suggest starting simple though, and if we're shrinking, then just try every time, and not bother to check or set that oom flag. (But leave the existing behavior for the growing case.)

What do you think?

job.c
```diff
@@ -72,6 +72,40 @@ rehash()
     }
 }
 
+static void
+rehash_down()
```
kr: This function is almost identical to rehash, it just chooses a different next table size (cur_prime-1 instead of cur_prime+1). Consider parameterizing the existing function instead of copy-pasting it.

Shir0kamii:
Hi,

I'm happy to have feedback on this!

As you can see, there's some commits that should belongs only to the fork of my company. Sorry for that, it seems I made some mistakes using git.

I'll work on your suggestions today so that you can review the changes if you still have time!

Shir0kamii:
Well, this should be good.

About the load factor, I misread your intentions in my first commits. I believed that an hashtable should always be almost empty. I understand that it's not true, but that it have consequences on performances. Some users could prefer using memory to improve performance, or the opposite. To make beanstalkd more flexible, I think the best is to provide two options `--load-factor-upscale` and `--load-factor-downscale` to let the user choose. What do you think about this ?

As for minimizing the calls to calloc, the implementation was already doing what you suggested : when calloc fails, I flag the current prime number and stop calling calloc when you're trying to grow beyong the flagged prime number. Is this what you meant or do you want me to reset the flagged number when we shrink ?

Thanks for the work!

silviucpp:
@kr anything missing from here in order to have this merged into the master?

job.c
```diff
@@ -34,24 +34,32 @@ store_job(job j)
...
+    }
 
-    all_jobs_cap = primes[++cur_prime];
+    all_jobs_cap = primes[cur_prime];
     all_jobs = calloc(all_jobs_cap, sizeof(job));
     if (!all_jobs) {
         twarnx("Failed to allocate %zu new hash buckets", all_jobs_cap);
-        hash_table_was_oom = 1;
-        --cur_prime;
+        hash_table_max_prime = cur_prime;
```
kr: What I said in https://github.com/kr/beanstalkd/pull/295#discussion_r58302039 still applies.

The name `hash_table_max_prime` is incorrect. There is no such thing as a "max prime" or a maximum size for the hashtable. This code branch is concerned with the *out of memory* condition, which can happen for many reasons, including allocation of entirely unrelated objects. This should be put back the way it was.

Additionally, now that we're also shrinking the table, we should clear the oom flag in the other code path. Below this if statement, there should be a line:

```
if (!all_jobs) {
    twarnx(...);
    hash_table_was_oom = 1;
    ...
}
hash_table_was_oom = 0;
```

job.c
```diff
@@ -34,24 +34,32 @@ store_job(job j)
...
-rehash()
+rehash(int is_upscaling)
 {
     job *old = all_jobs;
     size_t old_cap = all_jobs_cap, old_used = all_jobs_used, i;
+    int old_prime = cur_prime;
 
-    if (cur_prime >= NUM_PRIMES) return;
-    if (hash_table_was_oom) return;
+    if (is_upscaling) {
```
kr: Please consider defining an index delta and using that to find the next prime:

```
int d = is_upscaling ? 1 : -1;

if (cur_prime + d >= NUM_PRIMES) return;
if (cur_prime + d < 0) return;
if (is_upscaling && hash_table_was_oom) return;

cur_prime += d;
```

This will make the code a little easier to follow

(I also sneakily fixed an off-by-one error here that was present in the original code, in the comparison against NUM_PRIMES.)

Shir0kamii:
@kr I did what you suggested, I hope it is now ready to merge!

job.c
```diff
@@ -12,7 +12,7 @@ static job *all_jobs = all_jobs_init;
 static size_t all_jobs_cap = 12289; /* == primes[0] */
 static size_t all_jobs_used = 0;
 
-static int hash_table_was_oom = 0;
+static int hash_table_was_oom = 1;
```
kr: So close! Thank you for doing all these revisions. :)

We just need to initialize this variable to 0 here. (If it's initialized to 1, the hash table will never grow.)

Shir0kamii:
@kr Sorry about this mistake, it's fixed!

kr:
Thank you!

ultrabug:
Kindly requesting a tag & release for this please so we can package it ASAP !

Thanks everyone.

PR #296, opened 2016-02-10
use syslog as well as writing to stderr for warnings

Roguelazer:
It's a lot more convenient to have daemons log to syslog than print errors to stderr; this patch sends all of the `warn()` / `warnx()` / etc. logging to syslog, as well as adding a log line on startup.

### DIFF
--- a/main.c
+++ b/main.c
@@ -70,7 +70,7 @@ main(int argc, char **argv)
     if (r == -1) twarnx("make_server_socket()"), exit(111);
     srv.sock.fd = r;
 
-    syslog(LOG_DEBUG, "Starting up on %s:%d", srv.addr, srv.port);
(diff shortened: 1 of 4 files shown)


JensRantil:
A workaround would be to simply pipe the output to `syslog`. Also, many `init` systems will already provide logging the stdout/stderr to files on a machine.

Roguelazer:
Both of owe are options, but they can't do things like include the message priority in the syslog PRI field, and often require a second process.

-- 
James Brown, currently mobile

kr:
Thanks a bunch for working on this patch! Unfortunately, it doesn't really make sense for beanstalkd.

There are many logging facilities that people use in production. I don't think it's a good idea for beanstalkd to give special status to just one. They can all consume a text stream of log lines, which we provide on stdout. (All beanstalkd messages have the same "priority", so if you're using syslog, you might just as well use `logger -p` to put whatever value you want in that field.)

PR #301, opened 2016-04-02
MOD: w->next change to max in function walread

git-hulk:
hello @kr 

I think it's a good idea to use max to replace w->next in function walread.

```c
void
walread(Wal *w, job list, int min, int max)
{
     File *f;
     int i, fd;
     int err = 0;
     
     for (i = min; i < w->next; i++) {
         ...
     }
}
```

### DIFF
--- a/walg.c
+++ b/walg.c
@@ -439,13 +439,13 @@ waldirlock(Wal *w)
 
 
 void
-walread(Wal *w, job list, int min, int max)
+walread(Wal *w, job list, int min)
 {
     File *f;
     int i, fd;
     int err = 0;
 
-    for (i = min; i < max; i++) {
+    for (i = min; i < w->next; i++) {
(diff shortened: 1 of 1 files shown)


JensRantil:
> I think it's a good idea to use max to replace w->next in function walread.

Could you elaborate why?

kr:
Looks like max is unused in walread. I don't know what I was thinking when I wrote that. It takes the same value as w->next, but "max" is the wrong name, because it's actually one more than the max. Anyway, why don't we just delete that param.

git-hulk:
hi @JensRantil 

As @kr said, max is unused in walread, and it may make confused if we want to use max to control min and max in walread.

@kr  yeah, it's also a good idea to delete param max. But it's  not differece between wal->next and max in walread, and use max would be more flexible.

PR #232, opened 2014-03-05
Updated download.md - escapestudios repo changes

djoos:
Hi there,

just a quick update of the download.md to reflect recent Escape Studios repo changes...

Thanks in advance for merging the update!

Kind regards,
David

PR #306, opened 2016-04-03
Use wiki for deployment/automated installation tools

JensRantil:
I stumbled across #215 and it just felt like it would be more natural to have on the [Tools page](https://github.com/kr/beanstalkd/wiki/Tools). I've updated the wiki page to include the puppet and chef stuff. This pull request simply refers users there.

### DIFF
--- a/download.md
+++ b/download.md
@@ -90,19 +90,7 @@ Just type:
 
 ## Automated installation
 
-### Chef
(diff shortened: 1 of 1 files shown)


JensRantil:
#232 is also somewhat related.

PR #307, opened 2016-04-03
Add draining mode status to the `stats` command

JensRantil:
This supersedes, fixes

> The feature looks useful, but I think it would be more straightforward to put `draining: false` or `draining: true` in the output. This should be a 2-line change.

and closes #219. Had to create a separate pull request and not build on @justincase's because his repository didn't seem to exist anymore.

```
binlog-max-size: 10485760
id: 42cdbb04ec381e89
hostname: Mac.local
draining: false
```

```
$ kill -SIGUSR1 <pid>
```

```
binlog-max-size: 10485760
id: 42cdbb04ec381e89
hostname: Mac.local
draining: true
```

### DIFF
--- a/prot.c
+++ b/prot.c
@@ -886,7 +886,7 @@ fmt_stats(char *buf, size_t size, void *x)
 
     if (drain_mode)
     {
-        srv->draining = "true";
+        srv->drain_mode = "true";
     }
 
     getrusage(RUSAGE_SELF, &ru); /* don't care if it fails */
@@ -939,7 +939,7 @@ fmt_stats(char *buf, size_t size, void *x)
             srv->wal.filesize,
             id,
             node_info.nodename,
-            srv->draining);
+            srv->drain_mode);
 
 }
 
--- a/prot.c
+++ b/prot.c
@@ -187,6 +187,7 @@ size_t job_data_size_limit = JOB_DATA_SIZE_LIMIT_DEFAULT;
     "binlog-max-size: %d\n" \
     "id: %s\n" \
     "hostname: %s\n" \
+    "draining: %s\n" \
     "\r\n"
 
 #define STATS_TUBE_FMT "---\n" \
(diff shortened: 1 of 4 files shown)


prot.c
```diff
@@ -932,7 +938,8 @@ fmt_stats(char *buf, size_t size, void *x)
             srv->wal.nrec,
             srv->wal.filesize,
             id,
-            node_info.nodename);
+            node_info.nodename,
+            srv->draining);
```
kr: How about `drain_mode ? "true" : "false"` here instead?

And I suggest putting this before the id field instead of after, so there's less noise in the diff.

JensRantil: Fixed in 90950db6c378febdf12326898cfbba65def0cb23. Let me know if you need anything else here.

prot.c
```diff
@@ -932,7 +938,8 @@ fmt_stats(char *buf, size_t size, void *x)
             srv->wal.nrec,
             srv->wal.filesize,
             id,
-            node_info.nodename);
+            node_info.nodename,
+            srv->drain_mode);
```
kr: Apologies; I didn't explain my feedback clearly enough the first time.

> How about `drain_mode ? "true" : "false"` here instead?

This is a suggestion to use the expression here, on this line, instead of `srv->drain_mode`. We already store the information about whether we're in drain mode in the global variable drain_mode, so there's no need to store it also in a field on srv. In fact, storing it redundantly is extra complexity that increases the chances of a bug in the future. It's better to compute the value here, directly from the boolean flag.

One way to do that is with a ternary expression here on this line. It's not the only way, but I recommend it because it's simple and straightforward.

> And I suggest putting this before the id field instead of after, so there's less noise in the diff.

It seems like you chose to ignore this suggestion. If you disagree with it that's fine, I'd just like to know why. Is there a particular reason to put the drain mode last in the list? If not, it would be marginally better somewhere else.

Putting this all together, we would get:

```
...
srv->wal.nrec,
srv->wal.filesize,
id,
drain_mode ? "true" : "false",
node_info.nodename);
```

PR #309, opened 2016-04-03
Make `protocol.txt` nicer to the eye

### LINKED ISSUE #287: Protocol doc reference improvement proposal
I tend to go back to the [protocol doc](https://raw.githubusercontent.com/kr/beanstalkd/master/doc/protocol.txt) every now and then. Generally it's super well written, explaining the various job states etc.

Nowadays, I'm fairly confident about the overall structure of beanstalk. This means I mostly use the protocol document for reference. I guess you could call me a power user. However, every time I open up the document it nags me that it's not optimized to be used as a reference. More specifically, skimming through all the commands is hard answering questions such as

> I know there's a command for pausing a tube - what was it?

> Which order were the parameters of `put` in?

> Which stats commands are there?

I have no problem with ASCII per se, but the command examples simply blend in really well with the rest of the document. Inspired by Markdown, would you accept a pull request to the procol doc that indents all the protocol examples? We would go from

```
Worker Commands
---------------

A process that wants to consume jobs from the queue uses "reserve", "delete",
"release", and "bury". The first worker command, "reserve", looks like this:

reserve\r\n

Alternatively, you can specify a timeout as follows:

reserve-with-timeout <seconds>\r\n

This will return a newly-reserved job. If no job is available to be reserved,
beanstalkd will wait to send a response until one becomes available. Once a
job is reserved for the client, the client has limited time to run (TTR) the
job before the job times out. When the job times out, the server will put the
job back into the ready queue. Both the TTR and the actual time left can be
found in response to the stats-job command.
```
to
```
Worker Commands
---------------

A process that wants to consume jobs from the queue uses "reserve", "delete",
"release", and "bury". The first worker command, "reserve", looks like this:

    reserve\r\n

Alternatively, you can specify a timeout as follows:

    reserve-with-timeout <seconds>\r\n

This will return a newly-reserved job. If no job is available to be reserved,
beanstalkd will wait to send a response until one becomes available. Once a
job is reserved for the client, the client has limited time to run (TTR) the
job before the job times out. When the job times out, the server will put the
job back into the ready queue. Both the TTR and the actual time left can be
found in response to the stats-job command.
```

. What do you think?

JensRantil:
This fixes #287.

### DIFF
--- a/doc/protocol.txt
+++ b/doc/protocol.txt
@@ -185,7 +185,9 @@ Parameters:
 
 The only reply is:
 
-USING <tube>\r\n
+    USING <tube>\r\n
+
+Parameters:
 
  - <tube> is the name of the tube now being used.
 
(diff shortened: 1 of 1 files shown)


JensRantil:
@kr I noticed you did not include 52cb941c7c85dda2430652d9c9da672487ee8666 nor 680415d226c19ae251ba866a2965a8a41036ccc8. Any particular reason for that?

Issue #305, opened 2016-04-03
Missing updates in Chinese protocol document


JensRantil:
a92e3eb9a7b24df917669246815818e5d8efeb4d does not have an equivalent change in `doc/protocol.zh-CN.md`.

JensRantil:
Maybe @tdg5 could help fix this? ;)

PR #312, opened 2016-04-03
Document stats-job values delay and ttr in Chinese

tdg5:
Addresses #305, **maybe**. I don't speak Chinese, but this copy-pasta may be sufficient. Text sourced from [the arguments to the put command](https://github.com/kr/beanstalkd/blob/master/doc/protocol.zh-CN.md#put).

@JensRantil any ability to verify?

### DIFF
--- a/doc/protocol.zh-CN.md
+++ b/doc/protocol.zh-CN.md
@@ -300,6 +300,8 @@ data 为YAML file的统计信息
 - `state` 表示job的当前状态
 - `pri` 表示job的优先级
 - `age` 表示job创建的时间单位秒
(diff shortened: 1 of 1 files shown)


JensRantil:
> any ability to verify?

Unfortunately not :-/

tdg5:
According to translate.google.com:
> - `Delay` integer value delay in seconds ready, the job at this time is delayed state. 
> - `Ttr` - time to run - integer value, allowing the worker to perform the maximum number of seconds, during this time if the worker can not delete, release, bury job, then the job timeout, the server will release this job, this job the state transition is ready. The minimum is 1 second, if the client is specified as 0 will be reset to 1.

**`¯\_(ツ)_/¯`**

PR #318, opened 2016-04-19
Added support for SIGINT and SIGTERM

rtuk:


JensRantil:
How about, it first draining before shutting down when getting SIGINT?

kr:
Beanstalkd uses a [crash-only](https://www.usenix.org/legacy/events/hotos03/tech/full_papers/candea/candea.pdf) design. There's no need to treat these signals specially.

kr:
Thanks for the effort! If you're interested in contributing, maybe take a look at the list of [open bugs](https://github.com/kr/beanstalkd/issues?q=is%3Aopen+is%3Aissue+label%3Abug) and pick something?

kr:
I'm going to go ahead and close this, but please do consider looking at the open issues, or if you think this aspect of beanstalkd could be improved, we can totally discuss your proposal. Thanks again!

Issue #327, opened 2016-07-30
Should the number of jobs in the tube be limited?


omigafu:
Hi, our group put forward a new idea in using beanstalkd. What we have done is that we limit the numbers of jobs through parameter `-m` and already use it in our project. It can reduce the pressure on producers, consumers and disk. Is it necessary to merge it?
### Method

`-m`    max count of maintain job, format: `tube1:max_conut,tube2:max_count`
           if tube is empty,set to  all the othrer tubes
`-x`      set the max count of bury job
### example

`./beanstalkd -m test:10,:5 -x 3`  It indicates that the maximum number of the jobs of tube `test`is 5, the maximum number of the jobs of nonspecific tube is 5, while the maximum number of the buried jobs is 3.
`./beanstalkd -m :5 -x 3`  It indicates that the maximum number of the jobs of nonspecific tube is 5 and the maximum number of the buried jobs is 3.

fuxiaotong:
Hi, welcome to leave comments for this issue : )

thanks.

JensRantil:
> It can reduce the pressure on producers, consumers and disk. Is it necessary to merge it?

Could you expand a little bit on this? Why does it reduce the pressure on producers, consumers and disk?

silviucpp:
Yes I don't understand as well. If the load is to big on a node you add more beanstalkd nodes and spread the load .. I don't see how limiting the number of jobs can help.. anyway you need to send that jobs somewhere ...

Silviu

fuxiaotong:
Hi, @JensRantil @silviucpp @kr I am so sorry for the late reply.

For more urgent tube, we don't want to have too much jobs in this tube and don't want to affect the jobs that needs to be processed urgently, so I would rather discard excess jobs.

By the way, for those non-emergency tubes, we can set the defult configuration settings for those tubes which is no limited. 

In general, this is just a way for us to use beanstalkd, and it is the choice you can choose to use. 

xiaotong

ysmolsky:
This does feel right in terms of what beanstalkd does, and especially using command lines parameters to override the behavior of server so drastically. This would break protocol promises made for version 1. Right now this repo is maintenance mode meaning that we are going to keep the thing running but without introducing major features.

omigafu:
Hi @ysmolsky , thanks for your reply.

PR #330, opened 2016-08-03
limit number of jobs in tube

omigafu:
We limit the numbers of jobs through parameter `-m` and already use it in our project. It can reduce the pressure on producers, consumers and disk. Is it necessary to merge it?


fuxiaotong:
I have submitted a issue at #327 , welcome to leave comments for this PR. : )

PR #331, opened 2016-08-03
Add support for put-in-tube command.

silviucpp:
Many clients like `Pheanstalk`, `ebeanstalkd` offers this functionality by sending 2 commands : use and put .

From my tests this is decreasing the performances a lot.
Using this new command the put is done in one single request and has the same performances as put command.

Command format is :

```
   put-in-tube tube pri delay ttr length\r\n
   data\r\n
```

### DIFF
--- a/doc/protocol.txt
+++ b/doc/protocol.txt
@@ -188,6 +188,15 @@ The only reply is:
 
  - <tube> is the name of the tube now being used.
 
+The "put-in-tube" command is for any process that wants to insert a job into a specified tube.
+It comprises a command line followed by the job body:
+
+    put-in-tube <tube> <pri> <delay> <ttr> <bytes>\r\n
+    <data>\r\n 
+
+Works in the same way as "put" the only difference being that it inserts a job into the specified tube and not in
+the client's currently used tube (see the "use" command below).
+
 Worker Commands
 ---------------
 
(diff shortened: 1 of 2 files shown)


silviucpp:
I believe you are wrong and doesn't take in consideration many other situations like:

1. in case use command fails for some reason the put command will add the task in a wrong tube. You can take a look in redis and see that this sort of commands sent in pipeline where command 2 depends on 1 are sent in transactions (if one fails both fails. which is not supported by BK protocol). 

2. Lot of languages like java and erlang where you can use pool of connections can take advantage of this new command . Instead of getting a connection from pool set the tube you want and the push the event. You only take the connection and send one command only. 

3. Lot of languages like php (which is very popular) from my knowledge won't be able to use pipeline. 

The command is pretty simple and implementation straight forward and I don't see how BK can improve without new features.

silviucpp:
I don't want to go into details but what you are saying I'm sure you can't find in a code wrote in professional solutions. If I run SELECT 1 to check if a connection is alive probably I'll be fired in the next second. There are many ways to check if a connection is alive without sending additional queries to a server. 

I perfectly understand your pipeline idea but again if you check the code USE command can fail for lot of reasons. and if that one fails the put is executed in a wrong tube. If you believe that code is perfect and cannot fail than your approach is ok even if USE still doing allocations/deallocations and all kind of parsing which is avoided on the server side using one single command.

I'm not interested in merging the code as time I'm maintaining my own fork, but will all the respect with this approach where you are closing peoples pull requests without coming with strong arguments this project will never come back to life.

PR #334, opened 2016-09-22
Handle jobs not being released when they expire

sanjeethr:
prottick invokes conn_timeout periodically to handle connections
having jobs that have expired and release them back to ready.
But when conn_timeout is invoked on a connection which has no
jobs expiring it currently leaks this connection and associated
jobs since the connection is removed from the server's connection
heap. If the client hangs and the connection is kept open, these
jobs are leaked till the client is restarted.

The fix is to call connsched irrespective of whether any jobs
were reclaimed or not, instead of calling it for every job
reclaimed. One fundamental change is since connsched is not
called after every job is released, rebalancing of the connections
heap does not happen every time. Instead, it happens only once
post all jobs being handled in the loop. I do not think this is an 
issue. Please let me know otherwise.

Resolves : https://github.com/kr/beanstalkd/issues/262

I have run  make check and make bench and attached the results.
Also, I have run this with some of our applications at decent scale for
testing and it worked fine
[benchmarks.txt](https://github.com/kr/beanstalkd/files/486511/benchmarks.txt)


meox:
please resync this branch

ysmolsky:
> But when conn_timeout is invoked on a connection which has no
> jobs expiring it currently leaks this connection and associated
> jobs since the connection is removed from the server's connection
> heap. If the client hangs and the connection is kept open, these
> jobs are leaked till the client is restarted.

This is not true according to how I model the code execution. Connection is not removed from the heap when it has reserved jobs, it is rescheduled.

I would like to see a test that proves me wrong. This test also could be a regression test showing that this PR actually solves the problem.

kr:
+1 for being careful here. Function conn_timeout is one of the trickier parts of beanstalkd. In the past I tried to make this logic easier to understand (so easier to make changes to it safely) but I didn't get very far.

ysmolsky:
I am closing because it won't be merged and I cannot propose suggestion to this PR to fix that.

Issue #339, opened 2017-02-17
Makefile: make it cross-compilation friendly


pprindeville:
I'm porting this package to OpenWRT.

I've noticed a few things...  (1) `rawfalloc()` should call `ftruncate()` instead as this is more portable than `posix_fallocate()`, and (2) it's not uncommon (in a cross-compile) for `PREFIX`, `BINDIR`, `CFLAGS`, `LDFLAGS`, `GCC`, and `LD`, to be externally seeded...

So, rather than setting `CFLAGS` as:

```
CFLAGS=-Wall -Werror\
       -Wformat=2\
       -g\

```

I'd supplement `CFLAGS` as:

```
CFLAGS += -Wall -Werror -Wformat=2 -g
```

and ditto for `LDFLAGS`, etc.  Assume that `PREFIX` might also be set, i.e.:

```
PREFIX ?= /usr/local
```

ditto for `INSTALL`, `BINDIR`, etc.

See PR #340.

pprindeville:
What's the schedule for the next release, btw?

sergeyklay:
@pprindeville https://github.com/kr/beanstalkd/issues/299

pprindeville:
@sergeyklay: So...  I should be using https://github.com/silviucpp/beanstalkd-deb instead?

sergeyklay:
I have no idea :)

silviucpp:
Hello guys,

The mention project is only a way to create yourself a DEB that can be installed on ubuntu using the last beanstalkd master.  In case you want another revision or fork you only need to change the properties in 
https://github.com/silviucpp/beanstalkd-deb/blob/master/setup/setup.properties

This anyway doesn't solve the problem that beanstalkd is not longer an active maintain project.

Silviu

ysmolsky:
@pprindeville I totally support this initiative. Would you be so kind to update your PR because we have changed Makefile recently?

pprindeville:
> I totally support this initiative. Would you be so kind to update your PR because we have changed Makefile recently?

@ysmolsky Done.

PR #340, opened 2017-02-17
Have a more cross-compilation friendly Makefile

pprindeville:
In the OpenWRT, builds are often done under one system (like x86
Ubuntu) for an ARM image of OpenWRT.  CFLAGS, GCC, LDFLAGS, LD,
DESTDIR, PREFIX, etc. all need to be supplied for the target
system.

Addresses issue #339.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,11 +1,13 @@
-PREFIX=/usr/local
+PREFIX?=/usr/local
 BINDIR=$(DESTDIR)$(PREFIX)/bin
 
 override CFLAGS+=-Wall -Werror -Wformat=2 -g
 override LDFLAGS?=
 
-OS=$(shell uname|tr A-Z a-z)
-INSTALL=install
+LDLIBS?=
+
+OS?=$(shell uname | tr 'A-Z' 'a-z')
+INSTALL?=install
 
 VERS=$(shell ./vers.sh)
 TARG=beanstalkd
@@ -39,16 +41,12 @@ HFILES=\
 	sd-daemon.h\
 
 ifeq ($(OS),linux)
-
-LDLIBS=\
-	-lrt\
-
+   LDLIBS+=-lrt
 endif
 
 CLEANFILES=\
 	vers.c\
-	*.gcda\
-	*.gcno\
+	$(wildcard *.gcda *.gcno)
 
 .PHONY: all
 all: $(TARG)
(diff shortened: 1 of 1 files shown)


pprindeville:
@kr Can I get an up/down vote on whether this will be merged, please?

kr:
Just fyi, the existing makefile already lets you override variables in the standard way:

```
$ cat Makefile 
X=foo

default:
	echo $(X)
$ make
echo foo
foo
$ make X=bar
echo bar
bar
```

Would this be sufficient for your needs?

pprindeville:
It’s more about appending values to what might get passed in from the command-line, like $(CFLAGS) for instance.

pprindeville:
Got rid of the whitespace only changes so it was more obvious what was changing.

Also, used `$(wildcard ...)` to do globbing inside of Make, so that `make -n` shows more accurately what it will do.

Makefile
```diff
@@ -89,15 +82,18 @@ ct/ct.o ct/_ctcheck.o: ct/ct.h ct/internal.h
...
-endif
 vers.c:
 	./verc.sh >vers.c
 
 doc/beanstalkd.1 doc/beanstalkd.1.html: doc/beanstalkd.ronn
 	ronn $<
 
 freebsd.o: darwin.c
+
+.PHONY: all install clean check bench
```
ysmolsky: I find it easier to have PHONY section near the target. Even thought it's more lines, it's easier to manage and understand.

Makefile
```diff
@@ -39,43 +41,34 @@ HFILES=\
...
 
-CLEANFILES:=$(CLEANFILES) $(TARG)
+CLEANFILES+=$(TARG)
 
 $(OFILES) $(MOFILE): $(HFILES)
 
-.PHONY: clean
 clean:
-	rm -f *.o $(CLEANFILES)
+	rm -f $(wildcard *.o) $(CLEANFILES)
```
ysmolsky: maybe move *.o to CLEANFILES too?

pprindeville: Done

Makefile
pprindeville: A lot of people find it clutter and prefer it at the end where it's easier to ignore.  But I've made the change.

ysmolsky: I see your point. But I want to keep it on the safe, conservative side for a while. I appreciate your feedback.

PR #341, opened 2017-02-17
Use ftruncate() as it's much more portable

pprindeville:
Embedded systems (such as OpenWRT) often have reduced functionality; when operating in such circumstances, using the most widely available API's (or subsets of API's) is key for portability.  ftruncate() is much more ubiquitous than posix_fallocate().

### DIFF
--- a/linux.c
+++ b/linux.c
@@ -1,4 +1,6 @@
 #define _XOPEN_SOURCE 600
+#include <unistd.h>
+#include <sys/types.h>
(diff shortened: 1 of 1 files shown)


pprindeville:
@kr Trivial 2-line fix.  Can I get an up/down vote on whether this will be merged, please?

JensRantil:
Makes sense. Thanks!

PR #342, opened 2017-02-17
Fix headers for MUSL compatibility

pprindeville:
The MUSL headers generally don't like being included as <sys/foo.h> when the preferred means is <foo.h> (example: time.h) and will complain loudly when this happens.  It's unnecessary noise, and it
potentially draws attention from real warnings that might otherwise be overlooked.

### DIFF
--- a/sd-daemon.c
+++ b/sd-daemon.c
@@ -32,7 +32,7 @@
 #include <sys/stat.h>
 #include <sys/socket.h>
 #include <sys/un.h>
-#include <sys/fcntl.h>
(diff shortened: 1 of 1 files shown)


pprindeville:
@kr Can I get an up/down on whether this will be merged please?  It's over a year old, but it's a one-liner.

JensRantil:
Fixed in cf95cbce4af653dcf64e6162f7d73e3a570becfb.

pprindeville:
Thanks!

PR #356, opened 2017-07-04
update protocal.zh-CN, fix markdown

laixintao:
There should be a space between hashtag and tital, otherwise github won't parse this as title. see the img below.

before
image

after
image


laixintao:
duplicated with https://github.com/kr/beanstalkd/pull/353#pullrequestreview-47793477

PR #367, opened 2017-11-26
Report bound port correctly

ChoHag:
When binding to port 0 beanstalk correctly binds to a random port but doesn't report which port it bound to, reporting 0 instead. This patch fixes that.


kr:
Nice, thanks!

PR #387, opened 2018-06-29
The version script doesn't work as is

pprindeville:
It's also more complicated than it needs to be.

Signed-off-by: Philip Prindeville <philipp@redfish-solutions.com>

### DIFF
--- a/verc.sh
+++ b/verc.sh
@@ -1,5 +1,14 @@
 #!/bin/sh
 
-printf 'const char version[] = "'
+echo_n()
+{
+  if [ "`echo -n`" = "-n" ]; then
+    echo "$@""\c"
+  else
+    echo -n "$@"
+  fi
+}
+
+echo_n 'const char version[] = "'
 ./vers.sh
-printf '";\n'
+echo '";'
--- a/vers.sh
+++ b/vers.sh
@@ -1,11 +1,20 @@
 #!/bin/sh
 
+echo_n()
+{
+  if [ "`echo -n`" = "-n" ]; then
+    echo "$@""\c"
+  else
+    echo -n "$@"
+  fi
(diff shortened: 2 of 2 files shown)


verc.sh
```diff
@@ -1,5 +1,3 @@
 #!/bin/sh
 
-printf 'const char version[] = "'
-./vers.sh
-printf '";\n'
+echo "const char version[] = \"`./vers.sh`\";"
```
JensRantil: I don't think the above is simple. However, I do think
```
echo -n 'const char version[] = "'
./vers.sh
echo ';'
```
is.

pprindeville:
@JensRantil Please re-review.

kr:
Reading the diff and comments above, it’s not entirely clear to me what the problem is.

Do you think you can briefly describe:

• steps to reproduce the problem
• the behavior you saw
• the behavior you expected

Note that the printf shell builtin is more portable than echo. As you’ve seen, echo has different usage depending on the environment. Using printf instead keeps things simpler, because there’s no need for a conditional check.

Of course it’s possible these scripts are using printf incorrectly. In that case, it’s probably easier to fix the printf usage rather than replacing it with echo.

ysmolsky:
I second @kr opinion, it is not clear what is broken with printf. An example would be great.

pprindeville:
> Reading the diff and comments above, it’s not entirely clear to me what the problem is.

@kr Sorry, it was a year ago that I was having this issue.  I can't remember if the problems with building were on Ubuntu 16.05, MacOS 10.13, or Fedora 28...  I'll see if I can reproduce it.

ysmolsky:
Closing, please reopen if you can reproduce the issue.

PR #388, opened 2018-08-15
fix critical translation error

AmyLewis:


JensRantil:
Fixed in 227937f60b2f6f8499576c332eb5c2e1bc222630. Thanks!

PR #393, opened 2018-08-19
chore(adm): remove all init files

JensRantil:
Proposal here is to have this repository not focus on surrounding
infrastructure needed to run it in various environments; this means
_not_ maintaining

 * init system configuration files.
 * packaging stuff (RPM, Deb, Docker etc.).

**Why?** Two reasons:

 * Because it doesn't scale. There are lots of different system
   setups.
 * Because this helps us stay focused on the problem we are trying to
   solve, which is to have a badass and super useful small queue.

**So, where do we put packaging and system init configuration files?**
Let's maintain that in a separate repositories as long as someone can
take ownership of it. Just create an issue and ping me and I'll
create the repo.

This commit makes PR #178 & #175 obsolete.


pprindeville:
Disagree.  I work with packaging for a couple of other distros (OpenWrt, Fedora, CentOS, EPEL, etc) and upstream packaging is always preferred since changes upstream that then require different initialization don't always get noticed by the downstream packagers in a timely manner.

Also, why duplicate effort in N places downstream with each of the repos when it can be done by the best informed people once only upstream?

pprindeville:
Not sure anyone still use `upstart`, however.  Maybe that can safely be deleted.

ysmolsky:
I am not sure if this a good step. So I vote no.
@tomponline @sergeyklay opinions?

tomponline:
Agreed I prefer keeping them here.

sergeyklay:
I agree with idea to keep init files at the repo.

PR #394, opened 2018-08-19
doc(protocol): remove chinese version

JensRantil:
It's not scalable to require a translation of it every time we make
changes to the primary documentation, which is in English. If we want to
keep it we move it to a separate repository and make sure there is a
clear maintainer for it.


JensRantil:
@AmyLewis @miaoyushun @dkmccandless @mickeyouyou All of you have contributed to this document. Let me know if you'd like me to create a separate repository where you can maintain this translation.

ysmolsky:
@tomponline I am more on the side of excluding the Chinese translation from the repo. But I have no idea where to put it and how will maintain it?
@kr How did we end up with it in a first place? :)

sergeyklay:
There are nations that, for historical reasons very poorly understand English. For example the Chinese. For them, English is _a real problem_. I believe that if some enthusiast made this translation, this will ease the understanding of the project for Chinese readers a lot.

tomponline:
As the functionality of beanstalk is pretty stable (at this time)  it feels ashame to lose this useful documentation. Perhaps we could add a translated line at the top warning readers this could be out dated and request volunteers to help keep it updated. If in the future big changes in functionality or config occur we can reevaluate.

ysmolsky:
Alright, let's take a burden of keeping it then. I do not see how it can harm us.

PR #403, opened 2018-10-17
More safely closing connection

### LINKED ISSUE #328: conn.c: segfault in connclose
We're seeing an extremely rare segfault in 1.10. I don't have a core, but based on the EIP at the time of segfault, the error is happening at  https://github.com/kr/beanstalkd/blob/master/conn.c#L237 (`c->use` is null, so the deref is invalid and triggers a segfault).

I can't for the life of me figure out why `c->use` would be NULL. The only place it's explicitly set to NULL is on line 240, later in this same method.

This is the third time we've had beanstalkd segfault in the last year or so — it's just infrequent enough that I can't reproduce it by hammering beanstalkd with random data, but it's fairly catastrophic when it happens, so I also can't just ignore it and move on.

I am contemplating adding a guard (`if (c->use)`) before line 237 in our build, but that seems like it'd just be papering over the issue. I've seen this on multiple servers and multiple compiles of beanstalkd, so I'm fairly confident that it's not a cosmic ray or anything like that.

The segfault doesn't recur when beanstalkd restarts and loads the WAL, so it's doesn't seem to be a property of the data.

Here's what the logs end with:

```
Jul 30 05:51:13 $hostname beanstalkd[21487]: prot.c:1821 in update_conns: sockwant
Jul 30 05:51:13 $hostname beanstalkd[21487]: Operation not permitted
Jul 30 05:51:13 $hostname kernel: beanstalkd[21487]: segfault at 168 ip 0000000000401fa2 sp 00007ffde368d4b0 error 6 in beanstalkd[400000+e000]
```

It's all very strange.

sergeyklay:
This PR aims to resolve #328. Actually I prefer use branch prediction (see [__builtin_expect](https://kernelnewbies.org/FAQ/LikelyUnlikely)) but not sure if we can safely use it on all supported platforms.

### DIFF
--- a/conn.c
+++ b/conn.c
@@ -234,8 +234,10 @@ connclose(Conn *c)
     if (has_reserved_job(c)) enqueue_reserved_jobs(c);
 
     ms_clear(&c->watch);
-    c->use->using_ct--;
-    TUBE_ASSIGN(c->use, NULL);
+    if (c->use) {
+        c->use->using_ct--;
+        TUBE_ASSIGN(c->use, NULL);
+    }
 
     if (c->tickpos > -1) {
         heapremove(&c->srv->conns, c->tickpos);


sergeyklay:
@ysmolsky Could you please take a look

ysmolsky:
Is there any way to reproduce the error? I do not see stacktraces leading to this line in the relevant issue. Am I missing something? I would rather not to fix it without understanding what's happening.

pprindeville:
> Actually I prefer use branch prediction (see __builtin_expect) but not sure if we can safely use it on all supported platforms.

@sergeyklay Given that this only happens on teardown, do we really need to worry that much about a micro-optimization?

sergeyklay:
@pprindeville No need, you're right.

sergeyklay:
@ysmolsky Should I close this one

ysmolsky:
> @ysmolsky Should I close this one

I guess so. I would like to have more thorough investigation for crashes to see the path to resolution clearly.

ysmolsky:
Sergey, I am closing this for now. We might want dig deeper...

PR #414, opened 2019-06-21
add a code of conduct

kr:
Should have done this long, long ago.

### DIFF
--- a/CodeOfConduct.txt
+++ b/CodeOfConduct.txt
@@ -1,0 +1,78 @@
+Contributor Covenant Code of Conduct
+
+Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as
+contributors and maintainers pledge to making participation in our project and
+our community a harassment-free experience for everyone, regardless of age, body
+size, disability, ethnicity, sex characteristics, gender identity and expression,
+level of experience, education, socio-economic status, nationality, personal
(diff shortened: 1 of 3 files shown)


kr:
Hey folks. This project should have a code of conduct. The one at https://www.contributor-covenant.org/ seems pretty good.

Note that this isn't something I can just add a commit for on my own. We need project members to agree that it's a good thing, and it requires an ongoing commitment from people to enforce the rules. In particular, we need an email address to put in the document:

> Instances of abusive, harassing, or otherwise unacceptable behavior may be
reported by contacting the project team at [INSERT EMAIL ADDRESS].

If we have consensus here, let's decide what email address to put in the document, and I'll update this PR and we can merge it!

cc @JensRantil @tomponline @ysmolsky

ysmolsky:
Yes, this is something that is a must, so I vote for this step. As to email, I suggest, we setup google group (or another no-google alternative) where we invite all the members of the org. Its email address can be published in the code of conduct.

I see that we have this group: https://groups.google.com/forum/#!forum/beanstalk-talk, that is public, we could setup something like beanstalk-team that will be private.

tomponline:
Yep I am happy with this too.

ysmolsky:
Created a private group. Invited @JensRantil, @kr and @tomponline . We can use this email: groupbeanstalk-team@googlegroups.com

PR #425, opened 2019-06-28
Run benchmark on Travis CI

sergeyklay:


.travis.yml
```diff
@@ -30,6 +30,17 @@ script:
     - make $MAKEJOBS || ( echo "Build failure. Verbose build follows." && make V=1 ; false )
     - make check $MAKEJOBS VERBOSE=1
 
+jobs:
+  include:
+    - stage: "Benchmark: Linux + GCC"
+      os: linux
```
ysmolsky: We can run them on osx too. But I am not sure how we would use the output except that knowing that they were run without errors.

sergeyklay:
As a far as I know we can setup Travis in the future so that it will post message commit with bencmark results back to a pull request.

.travis.yml
sergeyklay: As a far as I know we can setup Travis in the future so that it will post message commit with bencmark results back to a pull request.

Let me add osx a bit later. I'm not at home now

sergeyklay:
@ysmolsky Added osx + clang. Could you define an environment variable in repository settings so that I'll try to deal with benchmark reports. For more see: https://damien.pobel.fr/post/github-api-from-travisci/

ysmolsky:
I have added GITHUB_TOKEN with repo/repo_public scope.

PR #431, opened 2019-06-30
test without forking to get accurate coverage

ysmolsky:


kr:
Note that this change doesn't prevent forking, it just limits the number of tests to one at a time.

kr:
The two fork calls I know of of the top of my head are:

- https://github.com/beanstalkd/beanstalkd/blob/5c01de1acdc4129bbfc74e875dd67fdd09370436/ct/ct.c#L197
- https://github.com/beanstalkd/beanstalkd/blob/5c01de1acdc4129bbfc74e875dd67fdd09370436/testserv.c#L43

I do not expect it to be trivial to remove either of them.

PR #437, opened 2019-07-03
improve readability of the code in some places

ysmolsky:
This is a start in a series of patches to fix
warnings from "cppcheck --enable=all --inconclusive --std=posix ."

### DIFF
--- a/file.c
+++ b/file.c
@@ -423,7 +423,7 @@ warnpos(File *f, int adj, char *fmt, ...)
     va_list ap;
 
     off = lseek(f->fd, 0, SEEK_CUR);
-    fprintf(stderr, "%s:%u: ", f->path, off+adj);
+    fprintf(stderr, "%s:%d: ", f->path, off+adj);
     va_start(ap, fmt);
     vfprintf(stderr, fmt, ap);
     va_end(ap);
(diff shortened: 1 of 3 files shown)


file.c
```diff
@@ -423,7 +423,7 @@ warnpos(File *f, int adj, char *fmt, ...)
     va_list ap;
 
     off = lseek(f->fd, 0, SEEK_CUR);
-    fprintf(stderr, "%s:%u: ", f->path, off+adj);
+    fprintf(stderr, "%s:%d: ", f->path, off+adj);
```
tomponline: @ysmolsky do you know why the data type changed here?

prot.c
```diff
@@ -435,7 +435,7 @@ process_queue()
 static job
 delay_q_peek()
 {
-    int i;
+    size_t i;
```
tomponline: @ysmolsky any possible issues with changing this type?

file.c
ysmolsky: The type of "off+adj" was signed int, so we just use the correct formatting directive to avoid possible confusion.

prot.c
ysmolsky: As you can see "i" is used for accessing by index. It's increments from 0 and never goes into negative. size_t is exactly the type that should be used for that matter. Since we never write i to files we are okay to change it.

ysmolsky: And most important: `    for (i = 0; i < tubes.used; i++) {`

tubes.used is size_t, meaning that if ever tubes.used reaches value > max_int, then there would be an overflow in i.

Issue #424, opened 2019-06-28
improve tests coverage


ysmolsky:
[![codecov](https://codecov.io/gh/beanstalkd/beanstalkd/branch/master/graph/badge.svg)](https://codecov.io/gh/beanstalkd/beanstalkd)

Right now tests coverage is at 32%. What is worse, some files are not tested at all:
conn.c, file.c, linux.c, prot.c, walg.c

https://codecov.io/gh/beanstalkd/beanstalkd/list/master/

We might want to pick low hanging fruit by adding integration tests that cover wide area.

kr:
The tests in testserv.c are mostly integration-style tests. They run srvserve in a subprocess and talk to it over a TCP socket. Since the code being tested is in another process, maybe the coverage tool can't see it.

sergeyklay:
The coverage tool reads `*.gcda` and `*.gcno` files which are created independent of the tool we're use to read them.

ysmolsky:
@sergeyklay not sure if I understood your statement. 

I looked into testsrv.c I see that server is started as fork and the test code is built coverage supported. Am I correct by saying that it does not matter how we start program that is compiled with coverage options because it will write instrumentation anyway?

sergeyklay:
@ysmolsky  Yes, I think so.

ysmolsky:
I think that our coverage is off because I found this: 
> If you want to perform test coverage analysis using gcov please be aware that gcov is not necessarily multi-process-safe. If you get strange coverage data, try -j1 and avoid forking in your test cases.

Source: https://github.com/kr/ct

ysmolsky:
Also we need to exclude some dirs from codecov with this: https://docs.codecov.io/docs/ignoring-paths

sergeyklay:
I'm totally missed this. Thank you. Good catch! About excluding dirs: which dirs do you want to exclude?

ysmolsky:
Most importantly, `ct`, also for clarity sake: `adm`, `pkg`, since we are not going to test them.

sergeyklay:
We should exclude test utils as well:
![image](https://user-images.githubusercontent.com/1256298/60700335-35075480-9f00-11e9-8d05-df96d16b534e.png)

ysmolsky:
While working on testing functions in prot.c I have noticed that even though it is covered by tests at least somewhat, this file is reported as 0% coverage. That is completely wrong.

kr:
[...]

PR #440, opened 2019-07-05
Improved code coverage report

sergeyklay:
Remove files and directories from coverage report matching exclude patterns:
- `ct/*`
- `adm/*`
- `pkg/*`
- `testheap.c`
- `testjobs.c`
- `testserv.c`
- `testutil.c`

I have no idea why, but ignoring via `codecov.yml` does not work.

Refs: #424

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,11 @@ git:
 
 language: c
 
+addons:
+    apt:
+      packages:
+          - lcov
+
 compiler:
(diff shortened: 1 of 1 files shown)


ysmolsky:
> I have no idea why, but ignoring via `codecov.yml` does not work.

It seems that codecov is buggy in general. Sometimes they just dont generate reports.

PR #445, opened 2019-07-07
fix types to satisfy -Wextra

ysmolsky:
There is lot of work to be done to satisfy -Wextra.
This is the first step where I have eliminated all
errors related to the non-strict type coercion.

Reduced the scope of some variables.

While at it I have eliminated function "one_cmd" as making
code more complex that it has to be.


sergeyklay:
I like it!

PR #446, opened 2019-07-07
set input width of pri, delay and ttr params to uint32

### LINKED ISSUE #230: extend delay/ttr bit-width and specify the types in the protocol
In the protocol doc, bit-widths are unspecified delay, ttr, and job id's. It's important for a client to know how large of an integer will be needed to represent any potential job id the server could send, and if these wraparound. Additionally, it would be useful for maximum durations to be specified, to avoid the beanstalkd returning errors that should be trivially detectable in the client.

ysmolsky:
Before this fix if the value specified in
pri/delay/ttr parameters would equal to 2**32
then it would be messed up and set to zero because of overflow.
And in general, the behaviour was not defined strictly
because the uint type can take different width.

I have fixed the read_num function to parse integer
that fits into range 0 .. 2**32-1. Internally it parses
integer into uint64, but it checks if it fits into uint32.
So it will return success only in this case.

I have added multiple tests checking that behaviour.

Fixes #230

### DIFF
--- a/dat.h
+++ b/dat.h
@@ -57,6 +57,9 @@ typedef int(FAlloc)(int, int);
 #define URGENT_THRESHOLD 1024
 #define JOB_DATA_SIZE_LIMIT_DEFAULT ((1 << 16) - 1)
 
+/* Maximum value (uint32) allowed in pri, delay and ttr parameters */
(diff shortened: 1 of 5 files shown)


tomponline:
Great to fix this 😃

PR #448, opened 2019-07-08
fix coverage for forked processes

### LINKED ISSUE #433: does forking in tests affect correctness of coverage repors?
@kr said in #431: "Note that this change doesn't prevent forking, it just limits the number of tests to one at a time." and

> The two fork calls I know of of the top of my head are:
> 
> * https://github.com/beanstalkd/beanstalkd/blob/5c01de1acdc4129bbfc74e875dd67fdd09370436/ct/ct.c#L197
> * https://github.com/beanstalkd/beanstalkd/blob/5c01de1acdc4129bbfc74e875dd67fdd09370436/testserv.c#L43
> 
> I do not expect it to be trivial to remove either of them.

It means that we need to investigate if forking affects the coverage correctness.

ysmolsky:
Previously, the forked process with a server was killed by SIGKILL. 
This prevented the process from writing coverage data.

This patch enables the child processes to handle the SIGUSR2 signal.
testserv.c setups a handler that flushes coverage data prior exiting.
This handler is setup only when WITHCOV is defined in CFLAGS.

Coverage after the patch:

lines: 64.9% (1434 out of 2209)
branches: 50.9% (649 out of 1274)

Updates #424
Fixes #433

### DIFF
--- a/ct/ct.c
+++ b/ct/ct.c
@@ -167,7 +167,7 @@ waittest(Test *ts)
     if (pid == -1) {
         die(3, errno, "wait");
     }
-    killpg(pid, SIGUSR2);
+    killpg(pid, 9);
 
     for (t=ts; t->f; t++) {
         if (t->pid == pid) {
--- a/ct/ct.c
+++ b/ct/ct.c
@@ -167,7 +167,7 @@ waittest(Test *ts)
     if (pid == -1) {
         die(3, errno, "wait");
     }
-    killpg(pid, 9);
+    killpg(pid, SIGUSR2);
 
     for (t=ts; t->f; t++) {
         if (t->pid == pid) {
--- a/testserv.c
+++ b/testserv.c
@@ -86,19 +86,15 @@ mustdiallocal(int port)
     return fd;
 }
 
-#ifdef WITHCOV
-void __gcov_flush (void);
-
-void
-gcov_flush(int signum)
+static void
+exit_process(int signum)
 {
-    __gcov_flush(); /* dump coverage data  */
-    usleep(500000); /* .5s; time to write data */
-    exit(1);        /* some tests are stuck without it */
+    printf("exiting from child\n");
+    exit(0);
 }
 
 static void
-set_sig_usr2()
+set_sig_handler()
 {
     struct sigaction sa;
 
@@ -109,14 +105,31 @@ set_sig_usr2()
         exit(111);
     }
 
-    sa.sa_handler = gcov_flush;
-    r = sigaction(SIGUSR2, &sa, 0);
+    sa.sa_handler = exit_process;
+    r = sigaction(SIGTERM, &sa, 0);
     if (r == -1) {
-        twarn("sigaction(SIGUSR2)");
+        twarn("sigaction(SIGTERM)");
         exit(111);
     }
 }
-#endif
+
+static void
+kill_srv_atexit(void)
+{
+    /* if we are already in child do not exit.
+       Code below can make tests pass,
+       but decreases coverage from 58.5 to 46.2%. */
+    /* if (srvpid != 0) */
+    /*     return; */
+
+    /* printf statements are for debugging purposes */
+    printf("killing %d\n", srvpid);
+    kill(srvpid, SIGTERM);
+
+    printf("waiting %d\n", srvpid);
+    waitpid(srvpid, 0, 0);
+    printf("exited %d\n", srvpid);
+}
 
 #define SERVER() (progname=__func__, mustforksrv())
 
@@ -146,6 +159,9 @@ mustforksrv()
     }
 
     if (srvpid > 0) {
+        /* srvpid is of a child. */
+        /* When the parent (test) has finished it will send sigterm to the child */
+        atexit(kill_srv_atexit);
         printf("start server port=%d pid=%d\n", port, srvpid);
         return port;
     }
@@ -153,9 +169,7 @@ mustforksrv()
     /* now in child */
 
     prot_init();
-#ifdef WITHCOV
-    set_sig_usr2();
-#endif
+    set_sig_handler();
 
     if (srv.wal.use) {
         struct job list = {};
@@ -700,7 +714,7 @@ cttest_binlog_empty_exit()
 
     port = SERVER();
 
-    kill(srvpid, 9);
+    kill(srvpid, SIGTERM);
     waitpid(srvpid, NULL, 0);
 
     port = SERVER();
@@ -742,7 +756,7 @@ cttest_binlog_basic()
     mustsend(fd, "\r\n");
     ckresp(fd, "INSERTED 1\r\n");
 
-    kill(srvpid, 9);
+    kill(srvpid, SIGTERM);
     waitpid(srvpid, NULL, 0);
 
     port = SERVER();
@@ -836,7 +850,7 @@ cttest_binlog_read()
     mustsend(fd, "delete 2\r\n");
     ckresp(fd, "DELETED\r\n");
 
-    kill(srvpid, 9);
+    kill(srvpid, SIGTERM);
     waitpid(srvpid, NULL, 0);
 
     port = SERVER();
--- a/testserv.c
+++ b/testserv.c
@@ -60,19 +60,18 @@ muststart(char *a0, char *a1, char *a2, char *a3, char *a4)
 static int
 mustdiallocal(int port)
 {
-    int r, fd;
     struct sockaddr_in addr = {};
 
     addr.sin_family = AF_INET;
     addr.sin_port = htons(port);
-    r = inet_aton("127.0.0.1", &addr.sin_addr);
+    int r = inet_aton("127.0.0.1", &addr.sin_addr);
     if (!r) {
         errno = EINVAL;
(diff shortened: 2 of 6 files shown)


kr:
Can I suggest using `SIGTERM` here instead, since we want the process to terminate.

testserv.c
```diff
@@ -87,12 +86,43 @@ mustdiallocal(int port)
...
 }
 
-#define SERVER() (progname = __func__, mustforksrv())
+#ifdef WITHCOV
+void __gcov_flush (void);
+
+void
+gcov_flush(int signum)
+{
+    __gcov_flush(); /* dump coverage data  */
```
kr: No need to call `__gcov_flush` here. Instead we can call `exit` and let gcov's atexit handler flush.

It's better not to call gcov internals directly if we can avoid it, and in this case we can.

testserv.c
```diff
@@ -87,12 +86,43 @@ mustdiallocal(int port)
...
+    usleep(500000); /* .5s; time to write data */
+    exit(1);        /* some tests are stuck without it */
```
kr: Why not `exit(0)`?

testserv.c
```diff
@@ -202,18 +236,14 @@ readline(int fd)
 static void
 ckresp(int fd, char *exp)
 {
-    char *line;
-
-    line = readline(fd);
+    char *line = readline(fd);
```
kr: There are quite a few unrelated changes in this PR, such as the one on this line. What do you think about making those separately?

testserv.c
```diff
@@ -123,6 +153,9 @@ mustforksrv()
     /* now in child */
 
     prot_init();
+#ifdef WITHCOV
```
kr: I think it's reasonable to install the signal handler unconditionally, especially if it uses `SIGTERM`.

ct/ct.c
```diff
@@ -167,7 +167,7 @@ waittest(Test *ts)
     if (pid == -1) {
         die(3, errno, "wait");
     }
-    killpg(pid, 9);
+    killpg(pid, SIGUSR2);
```
kr: I'd suggest `SIGTERM` here, since this is a pretty ordinary "kill the process, but let it do something before it exits" situation.

kr: Considering https://github.com/beanstalkd/beanstalkd/issues/424#issuecomment-509476476, maybe we could avoid making any patch to ct here. That might be preferable, since it'll make updating ct easier in the future.

ysmolsky:
@kr thanks for the help. I will look into your comments and adopt suggestions.

testserv.c
ysmolsky: I agree. I tend to add additional style improvements but i think it makes sense to not touch it here.

testserv.c
ysmolsky: I was meaning to use SIGTERM too because I think that this case should be handled as generally as possible.

ct/ct.c
ysmolsky: I have tried this approach with atexit, but it leads to some tests failed. These tests kill child-server in the middle of test and then the controlling wrapper thinks that since the test has received SIGTERM (15), it has failed:
```
ct/_ctcheck
................E...EE........................................

cttest_binlog_basic: error (signal 15)
start server port=55911 pid=86446
>20 put 0 0 100 0
>20
<20 INSERTED 1
start server port=55913 pid=86447
>22 delete 1
<22 DELETED

cttest_binlog_empty_exit: error (signal 15)
start server port=55921 pid=86455
start server port=55922 pid=86456
>25 put 0 0 0 0
>25
<25 INSERTED 1

cttest_binlog_read: error (signal 15)
start server port=55924 pid=86458
>25 use test
<25 USING test
>25 put 0 0 120 4
>25 test
<25 INSERTED 1
>25 put 0 0 120 4
>25 tes1
<25 INSERTED 2
>25 watch test
<25 WATCHING 2
>25 reserve
<25 RESERVED 1 4
<25 test
>25 release 1 1 1
<25 RELEASED
>25 reserve
<25 RESERVED 2 4
<25 tes1
>25 delete 2
<25 DELETED
start server port=55926 pid=86459
>27 watch test
<27 WATCHING 2
>27 reserve
<27 RESERVED 1 4
<27 test
>27 delete 1
<27 DELETED
>27 delete 2
<27 NOT_FOUND

0 failures; 3 errors.
make: *** [check] Error 1
lines: 62.3% (1392 out of 2236)
```

Alternatively,

We can skip everything in this patch if we change `ct.c` to `killpg(pid, SIGTERM)`. Or did the SIGKILL (9) exist because of cleanup reasons? Then I tried this, all tests were cleanup properly. Did you have something in mind when you used SIGKILL (9) in ct.c?

ysmolsky: > We can skip everything in this patch if we change ct.c to killpg(pid, SIGTERM). Or did the SIGKILL (9) exist because of cleanup reasons? Then I tried this, all tests were cleanup properly. Did you have something in mind when you used SIGKILL (9) in ct.c?

I take it back. It was the uncleaned cache of coverage data that made me to think that. It is not enough to change to SIGTERM.

ysmolsky:
@kr I have stuck with the solution proposed. Please see tests output, to which I have added some debugging prints. I am confused how the tests are run when we make two calls to SERVER() in one test. I guess some fork there does not do what is expected by testing framework. Please look at this and let me know. I am puzzled. I do want to solve it your way.

Output: https://travis-ci.org/beanstalkd/beanstalkd/jobs/556449326
```
ct/_ctcheck
................E...EE........................................
cttest_binlog_basic: error (signal 15)
start server port=46526 pid=5292
>22 put 0 0 100 0
>22 
<22 INSERTED 1
exiting from child
start server port=46084 pid=5293
>24 delete 1
<24 DELETED
exiting from child
killing 0
waiting 0
exited 0
cttest_binlog_empty_exit: error (signal 15)
start server port=42929 pid=5301
start server port=42870 pid=5302
>27 put 0 0 0 0
>27 
<27 INSERTED 1
start server port=42929 pid=5301
exiting from child
killing 0
waiting 0
exited 0
cttest_binlog_read: error (signal 15)
start server port=43778 pid=5304
>27 use test
<27 USING test
>27 put 0 0 120 4
>27 test
<27 INSERTED 1
>27 put 0 0 120 4
>27 tes1
<27 INSERTED 2
>27 watch test
<27 WATCHING 2
>27 reserve
<27 RESERVED 1 4
<27 test
>27 release 1 1 1
<27 RELEASED
>27 reserve
<27 RESERVED 2 4
<27 tes1
>27 delete 2
<27 DELETED
exiting from child
start server port=45984 pid=5305
>29 watch test
<29 WATCHING 2
>29 reserve
<29 RESERVED 1 4
<29 test
>29 delete 1
<29 DELETED
>29 delete 2
<29 NOT_FOUND
exiting from child
killing 0
waiting 0
exited 0
0 failures; 3 errors.
```

ysmolsky:
@kr I have uploaded the corresponding code and another commit, you might want to look at it.

ysmolsky:
I think I got it. After debugging I see that first fork in SERVER() makes some weird not needed atexit kill. Since the first SERVER() fork is being explicitly killed in the test, we can omit setting up atexit in first SERVER(), but in the second SERVER() we should keep it. Preparing the patch...

testserv.c
ysmolsky: Done

testserv.c
ysmolsky: yes, 0 is good enough.

ysmolsky:
Patch is done. @kr, please take a look.

testserv.c
```diff
@@ -670,9 +705,9 @@ cttest_binlog_empty_exit()
     srv.wal.use = 1;
     job_data_size_limit = 10;
 
-    port = SERVER();
+    port = SERVER_NO_ATEXIT();
 
-    kill(srvpid, 9);
+    kill(srvpid, SIGTERM);
```
kr: Good catch. 👍

testserv.c
```diff
@@ -87,12 +87,47 @@ mustdiallocal(int port)
...
+// to write gcov data to the filesystem before ct kills it with SIGKILL.
+static void
+kill_srv_atexit(void)
+{
+    kill(srvpid, SIGTERM);
+    waitpid(srvpid, 0, 0);
+}
+
+#define SERVER() (progname=__func__, mustforksrv(true))
+#define SERVER_NO_ATEXIT() (progname=__func__, mustforksrv(false))
```
kr: I wonder if it would be any better to test if srvpid is nonzero in kill_srv_atexit, so that we can install the atexit function unconditionally.

```c
static void
kill_srv_atexit(void)
{
    if (srvpid == 0) {
        return;
    }
    kill(srvpid, SIGTERM);
    waitpid(srvpid, 0, 0);
    srvpid = 0;
}
```

I think this logic would make it safe & okay for kill_srv_atexit to be registered two or more times.

(In that case it might also make sense to extract those last three lines into a helper function so it can be used in tests like cttestbinlogemptyexit.)

Just a thought. The `SERVER_NO_ATEXIT` approach is okay with me too.

ysmolsky: Very good observation!

Issue #443, opened 2019-07-05
Makefile: add -Wextra flag and fix errors


ysmolsky:
To make compiler more strict we can add `-Wextra` and fix these errors:
```
~/d/beanstalkd (master *%=) % make -k
cc -Wall -Werror -Wextra -Wformat=2 -g   -c -o conn.o conn.c
conn.c:17:13: error: unused parameter 'a' [-Werror,-Wunused-parameter]
on_watch(ms a, tube t, size_t i)
            ^
conn.c:17:31: error: unused parameter 'i' [-Werror,-Wunused-parameter]
on_watch(ms a, tube t, size_t i)
                              ^
conn.c:24:14: error: unused parameter 'a' [-Werror,-Wunused-parameter]
on_ignore(ms a, tube t, size_t i)
             ^
conn.c:24:32: error: unused parameter 'i' [-Werror,-Wunused-parameter]
on_ignore(ms a, tube t, size_t i)
                               ^
4 errors generated.
make: *** [conn.o] Error 1
cc -Wall -Werror -Wextra -Wformat=2 -g   -c -o file.o file.c
file.c:204:30: error: comparison of integers of different signs: 'int32' (aka 'int') and 'size_t' (aka 'unsigned long') [-Werror,-Wsign-compare]
            if (jr.body_size > job_data_size_limit) {
                ~~~~~~~~~~~~ ^ ~~~~~~~~~~~~~~~~~~~
file.c:328:30: error: comparison of integers of different signs: 'int32' (aka 'int') and 'size_t' (aka 'unsigned long') [-Werror,-Wsign-compare]
            if (jr.body_size > job_data_size_limit) {
                ~~~~~~~~~~~~ ^ ~~~~~~~~~~~~~~~~~~~
file.c:463:11: error: comparison of integers of different signs: 'int' and 'unsigned long' [-Werror,-Wsign-compare]
    if (n < sizeof(int)) {
        ~ ^ ~~~~~~~~~~~
3 errors generated.
make: *** [file.o] Error 1
cc -Wall -Werror -Wextra -Wformat=2 -g   -c -o heap.o heap.c
cc -Wall -Werror -Wextra -Wformat=2 -g   -c -o job.o job.c
cc -Wall -Werror -Wextra -Wformat=2 -g   -c -o ms.o ms.c
cc -Wall -Werror -Wextra -Wformat=2 -g   -c -o net.o net.c
[...]

PR #450, opened 2019-07-08
mark unused parameters to satisfy -Wextra

ysmolsky:
Parameters marked with the UNUSED_PARAMETERS macros cannot be
removed from functions because their prototypes cannot be changed.

Create the srv struct via designated initializers. Also remove NULL
assigment, because this way struct values are set to 0 anyway.

Also declare vars next to their first usage as usual.

Update #443

### DIFF
--- a/conn.c
+++ b/conn.c
@@ -16,6 +16,8 @@ int verbose = 0;
 static void
 on_watch(ms a, tube t, size_t i)
 {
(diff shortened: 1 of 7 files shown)


tomponline:
@ysmolsky looks good, are these compatible with clang btw?

ysmolsky:
Yes. Totally compatible. Sqlite does it too.

Issue #423, opened 2019-06-28
unable to build with optimizing levels on Linux


sergeyklay:
This error occurs only on turn full optimization eg `CFLAGS="-O3"`.

```bash
$ export CFLAGS="-O3"
$ export MAKEJOBS="-j$(getconf _NPROCESSORS_ONLN)"
```

```bash
$ make $MAKEJOBS
gcc -O3 -Wall -Werror -Wformat=2 -g   -c -o linux.o linux.c
gcc -O3 -Wall -Werror -Wformat=2 -g   -c -o conn.o conn.c
gcc -O3 -Wall -Werror -Wformat=2 -g   -c -o file.o file.c
file.c: In function ‘filewclose’:
file.c:536:9: error: ignoring return value of ‘ftruncate’, declared with attribute warn_unused_result [-Werror=unused-result]
         (void)ftruncate(f->fd, f->w->filesize - f->free);
         ^
cc1: all warnings being treated as errors
make: *** [file.o] Error 1
```

```bash
$ make bench
gcc -O3 -Wall -Werror -Wformat=2 -g   -c -o testheap.o testheap.c
gcc -O3 -Wall -Werror -Wformat=2 -g   -c -o testjobs.o testjobs.c
gcc -O3 -Wall -Werror -Wformat=2 -g   -c -o testserv.o testserv.c
gcc -O3 -Wall -Werror -Wformat=2 -g   -c -o testutil.o testutil.c
ct/gen testheap.o testjobs.o testserv.o testutil.o  >ct/_ctcheck.c.part
mv ct/_ctcheck.c.part ct/_ctcheck.c
gcc -O3 -Wall -Werror -Wformat=2 -g   -c -o ct/_ctcheck.o ct/_ctcheck.c
gcc -O3 -Wall -Werror -Wformat=2 -g   -c -o ct/ct.o ct/ct.c
ct/ct.c: In function ‘writetokens’:
ct/ct.c:543:18: error: ignoring return value of ‘write’, declared with attribute warn_unused_result [-Werror=unused-result]
             write(wjobfd, &c, 1); /* ignore error; nothing we can do anyway */
                  ^
ct/ct.c: In function ‘runbenchn’:
ct/ct.c:326:14: error: ignoring return value of ‘write’, declared with attribute warn_unused_result [-Werror=unused-result]
         write(durfd, &bdur, sizeof bdur);
              ^
[...]

PR #454, opened 2019-07-10
check the return value of ftruncate for possible error

ysmolsky:
This prevented compilation with -O2, -O3 levels on Linux with GCC.

Updates #423

### DIFF
--- a/file.c
+++ b/file.c
@@ -533,7 +533,10 @@ filewclose(File *f)
     if (!f) return;
     if (!f->iswopen) return;
     if (f->free) {
-        (void)ftruncate(f->fd, f->w->filesize - f->free);
+        errno = 0;
+        if (ftruncate(f->fd, f->w->filesize - f->free) != 0) {
+            twarn("ftruncate");
+        }
     }
     close(f->fd);
     f->iswopen = 0;


file.c
```diff
@@ -533,7 +533,11 @@ filewclose(File *f)
     if (!f) return;
     if (!f->iswopen) return;
     if (f->free) {
-        (void)ftruncate(f->fd, f->w->filesize - f->free);
+        errno = 0;
+        int r = ftruncate(f->fd, f->w->filesize - f->free);
+        if (r == -1) {
+            twarn("ftruncate failed, errno=%d", errno);
```
sergeyklay: Is f->path available here? Could we specify a file name here as well?

sergeyklay: btw, 
```c
if (ftruncate(f->fd, f->w->filesize - f->free) != 0) {
    errno = 0;
    twarn("ftruncate failed, errno=%d", errno);
}
```
a bit shorter and w/o extra variable :)

ysmolsky: You are right about messages, they should be improved but it should be done more systematically and as separate effort. So for now I will skip adding f->path.
Thanks.

kr: Note that twarn mostly just calls the stdlib function warn, which has errno/strerror built in.

```
The err(), errc(), verr(), verrc(), warn(), warnc(), vwarn(), and
vwarnc() functions append an error message obtained from strerror(3)
based on a supplied error code value or the global variable errno, pre-
ceded by another colon and space unless the fmt argument is NULL.
```

file.c
```diff
@@ -533,7 +533,10 @@ filewclose(File *f)
     if (!f) return;
     if (!f->iswopen) return;
     if (f->free) {
-        (void)ftruncate(f->fd, f->w->filesize - f->free);
+        errno = 0;
+        if (ftruncate(f->fd, f->w->filesize - f->free) != 0) {
+            twarn("ftruncate failed, errno=%d", errno);
```
kr: Note that twarn mostly just calls the stdlib function warn, which has errno/strerror built in.

```
The err(), errc(), verr(), verrc(), warn(), warnc(), vwarn(), and
vwarnc() functions append an error message obtained from strerror(3)
based on a supplied error code value or the global variable errno, pre-
ceded by another colon and space unless the fmt argument is NULL.
```

So I think this line would print about the same information if it were written like this:

```c
twarn("ftruncate"); 
```

ysmolsky: Thanks!

PR #455, opened 2019-07-10
compile beanstalkd in c99 mode by default

ysmolsky:
Some old compilers requires the -std=c99 parameter, so we can use
features of c99 without errors:
`‘for’ loop initial declarations are only allowed in C99 mode`

Failed build: https://travis-ci.org/beanstalkd/beanstalkd/jobs/556849984

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 PREFIX?=/usr/local
 BINDIR=$(DESTDIR)$(PREFIX)/bin
 
-override CFLAGS+=-Wall -Werror -Wformat=2 -g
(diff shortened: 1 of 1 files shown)


sergeyklay:
Hmmm :thinking: can we amend build matrix to be sure it works for both cases?

ysmolsky:
I think I will not go there yet. I am not ready to migrate this thing to C99 as a whole. So I will try to code that it won't require `-std=c99` for any [old] compilers. I think this is easier. Also I can live with this:
```
int i;
for (i=0; ...)
```
This was the only reason to think about adding `-std=c99` for now.

PR #456, opened 2019-07-11
fix the job size overflowing the int type and set hard limit on it

### LINKED ISSUE #401: segfault when job size is larger than INT_MAX=2147483648
Whenever client tries to put the job larger than INT_MAX beanstalkd is crashed with following stack trace:
```
(gdb) bt
#0  __memmove_ssse3_back () at ../sysdeps/x86_64/multiarch/memcpy-ssse3-back.S:97
#1  0x0000000000406583 in fill_extra_data (c=0x702040) at prot.c:796
#2  0x00000000004065de in _skip (c=0x702040, n=-1877792593, line=0x40e0e5 "JOB_TOO_BIG\r\n", len=13) at prot.c:808
#3  0x0000000000407b7b in dispatch_cmd (c=0x702040) at prot.c:1229
#4  0x0000000000409721 in do_cmd (c=0x702040) at prot.c:1660
#5  0x00000000004098ac in conn_data (c=0x702040) at prot.c:1700
#6  0x0000000000409e07 in h_conn (fd=18, which=114, c=0x702040) at prot.c:1842
#7  0x0000000000409eb7 in prothandle (c=0x702040, ev=114) at prot.c:1854
#8  0x000000000040b3f6 in srvserve (s=0x611520 <srv>) at serv.c:56
#9  0x000000000040d274 in main (argc=15, argv=0x7fffffffe468) at main.c:91
```
It happens because errno is never verified after `strtoul`. In this case `errno == ERANGE` and `body_size =  2417174701`. Since body_size is uint, its value does not fit in the n argument (int) of  the _skip function. It leads to overflow and in the _skip function `n = -1877792593`:
 https://github.com/beanstalkd/beanstalkd/blob/b5a6f7a23a368ffb31fbf48cdffe95541166d3fa/prot.c#L1221-L1230

I guess, beanstalkd should respond with BAD_FORM in the case when the job size does not fit into UINT_MAX. Also the n argument in _skip function should be of the uint type.

ysmolsky:
I have set the maximum allowed limit for the job size to 1024^3 bytes (1GB).
When user tries to specify (-z) the limit for the job size larger than that,
he will get a warning that limit was capped to 1GB.

The program stores body_size in int32 (see struct Jobrec) and verifies that 
parameter in the put command can be parsed and fit into int32. 
I have propagated changes to the type of all involved functions.

Fixes #401

### DIFF
--- a/dat.h
+++ b/dat.h
@@ -62,7 +62,12 @@ typedef int(FAlloc)(int, int);
 #define URGENT_THRESHOLD 1024
 #define JOB_DATA_SIZE_LIMIT_DEFAULT ((1 << 16) - 1)
 
-/* Maximum value (uint32) allowed in pri, delay and ttr parameters */
+// The maximum value that job_data_size_limit can be set to via "-z".
+// It could be up to INT32_MAX-2 (~2GB), but set it to 1024^3 (1GB).
+// The width is restricted by Jobrec.body_size that is int32.
+#define JOB_DATA_SIZE_LIMIT_MAX 1073741824
+
+// Maximum value (uint32) allowed in pri, delay and ttr parameters
 #define MAX_UINT32 4294967295
 
 #define UNUSED_PARAMETER(x) (void)(x)
@@ -127,7 +132,10 @@ enum // Jobrec.state
     Copy
 };
 
-// if you modify this struct, you must increment Walver above
+// If you modify this struct, you must increment Walver above,
+// Beanstalkd does not handle migration between different versions:
+// it will reject the old data and exit from the server.
+// TODO: Handle Walver migrations automatically.
 struct Jobrec {
     uint64 id;
     uint32 pri;
@@ -297,8 +305,8 @@ struct Conn {
     // while in_job_read is nonzero, we are in bit bucket mode and
     // in_job_read's meaning is inverted -- then it counts the bytes that
     // remain to be thrown away.
-    int in_job_read;
-    job in_job; // a job to be read from the client
+    int32 in_job_read;
+    job   in_job; // a job to be read from the client
 
     job out_job;
     int out_job_sent;
--- a/prot.c
+++ b/prot.c
@@ -774,14 +774,13 @@ which_cmd(Conn *c)
 static void
 fill_extra_data(Conn *c)
 {
-    int extra_bytes, job_data_bytes = 0, cmd_bytes;
-
     if (!c->sock.fd) return; /* the connection was closed */
     if (!c->cmd_len) return; /* we don't have a complete command */
 
     /* how many extra bytes did we read? */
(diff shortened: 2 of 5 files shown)


ysmolsky:
> @ysmolsky 1GB seems like a reasonable limit, as in some tests I've seen done previously beanstalkd slows down considerably with larger jobs. Is there a reason you selected 1GB though?

Ah, I have never explained it in the commit msg, only slightly hinted that `body_size` should fit int32. Here is the reason. `body_size` is stored in Jobrec that should not be changed often. it is int32 there, it can take values up to 2147483647 (slightly less than ~2GB). So since it cannot fit the value for 2GB I have picked the next low round value that is 1GB = 1024 ^ 3 bytes.
```
struct Jobrec {
...
    int32  body_size;
...
}
```
Should I add this explanation to the comment about `JOB_DATA_SIZE_LIMIT_MAX`?

tomponline:
@ysmolsky sounds good, it cant hurt :)

ysmolsky:
I do not understand why coverage has dropped so rapidly. I think something is wrong with linux gcc on travis, it's too slow or too restricted? @sergeyklay

sergeyklay:
@ysmolsky It is usual coverage side effect. 

Changing lines like
```c
if (c->in_job_read == 0)
    return reply(c, line, len, STATE_SENDWORD);
```

to
```c
if (c->in_job_read == 0) return reply(c, line, len, STATE_SENDWORD);
```

will increase code coverage metric.

Let's consider a small example. Obviously, this code will never be covered:
```diff
+ if (0) {
-     func1();
+ }
```

 However this will be:
```diff
+ if (0) func1();
```

ysmolsky:
> @ysmolsky It is usual coverage side effect.

I think there is something more to it here. It feels like some tests were not run at all, maybe due to increased memory consumption in newly added tests. This patch add two tests each consuming at least 1 Gb. Let me try to disable those two tests and see how it affects coverage.

ysmolsky:
The coverage generated by codecov is different from coverage generated on macOS locally. I have looked at prot.c. Some tests were not run in gcc-linux version! For example command kick was not tested in gcc-linux at all:
codecov:
Screen Shot 2019-07-15 at 22 52 33
locally:
Screen Shot 2019-07-15 at 22 52 45

How I "cover" on mac:
```
~/d/beanstalkd (jobsize=) % cat ~/bin/test-cover.sh
#!/bin/bash

make clean check -j1 \
     CFLAGS="-O0 -ggdb -fprofile-arcs -ftest-coverage" \
     LDFLAGS="--coverage"

gcovr \
    --exclude-directories ct \
    -e testheap.c \
    -e testjobs.c \
    -e testserv.c \
    -e testutil.c \
    -p \
    -r . \
    -s \
    --html-details -o cover.html
```

PR #457, opened 2019-07-11
fix struct initializers to fix -Wextra warnings

ysmolsky:
Updates #443

### DIFF
--- a/dat.h
+++ b/dat.h
@@ -17,6 +17,11 @@ typedef uint64_t      uint64;
 #define int64_t  do_not_use_int64_t
 #define uint64_t do_not_use_uint64_t
 
+/* TODO: typedefs of ms, job and tube should not hide the pointer.
+   Make them similar to other typedefs (Conn, Heap).
+   Maybem move each typedef next to the corresponding struct definition.
+   See issue #458. */
+
 typedef struct ms     *ms;
 typedef struct job    *job;
 typedef struct tube   *tube;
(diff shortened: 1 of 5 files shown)


ysmolsky:
It turned much harder to conform to -Wextra that I have expected. The only remaining fix is in ct now.
I am not sure why about how struct initializers took like now. They conform to all possible compilers that I could get my hand on, but I still not sure if that it the best way to initialize zero struct.

dat.h
```diff
@@ -17,6 +17,10 @@ typedef uint64_t      uint64;
 #define int64_t  do_not_use_int64_t
 #define uint64_t do_not_use_uint64_t
 
+/* WARNING: definitions of ms, job and tube are confusing.
+   They introduce types that are pointers. Probably,
+   they should be renamed to indicate their nature. */
+
```
sergeyklay: I vote for ms_ptr, job_ptr and so on. btw, how about TODO instead of WARNING?

kr: I agree it would be a good TODO to change these types. But I wouldn't introduce names like anything_ptr. Instead we can make these three types look like the others (e.g. Conn, Heap, File, etc), which are typedefs for struct types of the same names. I got this style from some Plan 9 programs, and as you can see I never got all the way through converting everything.

ysmolsky: I was meaning to make them look like the other type aliases, hiding pointer does not make it easy to read. Converting this into TODO

ysmolsky: Filed #458

PR #460, opened 2019-07-12
include dependent headers in the dat.h file

ysmolsky:
Before this fix my editor was hinting the `*_t` types as undefined. This was confusing.

I have rearranged dat.h and sd-daemon.h before other headers to verify this.
I have removed redefinitions of *_t types because it was producing redefinitions errors.

### DIFF
--- a/conn.c
+++ b/conn.c
@@ -1,11 +1,11 @@
+#include "dat.h"
+#include <errno.h>
+#include <limits.h>
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
-#include <errno.h>
-#include <limits.h>
 #include <unistd.h>
-#include "dat.h"
 
 #define SAFETY_MARGIN (1000000000) /* 1 second */
 
--- a/darwin.c
+++ b/darwin.c
@@ -1,3 +1,4 @@
+#include "dat.h"
 #include <stdint.h>
 #include <sys/types.h>
 #include <unistd.h>
@@ -6,7 +7,6 @@
 #include <errno.h>
 #include <sys/event.h>
 #include <sys/time.h>
-#include "dat.h"
 
 enum
 {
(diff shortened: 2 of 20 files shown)


kr:
This document describes a rule about include directives. I find it’s argument persuasive.

http://doc.cat-v.org/bell_labs/pikestyle

The rule is, C header files should not include other header files.

kr:
Also, the description of this PR seems inaccurate.

> remove dependency of dat.h on other headers

The dependency is still there, the difference is just that this PR includes the headers in dat.h itself.

ysmolsky:
Thanks for the reference. I might be a little uneducated here, so this helps.

ysmolsky:
Pike says that is to reduce the amount of parsing in the compiler. Sounds reasonable.  What about editors (emacs) that hints errors in dat.h for undefined types like size_t? I wanted to solve this particular case.

kr:
(Sorry for the terse comments above, I was on my phone and had spotty internet and I just wanted to get the basic message across while I could.)

The emacs thing is interesting. I wonder if there's a way to configure emacs to include the necessary files.

On the other hand, it might not be a problem for beanstalkd for dat.h to include system header files. Things could start to get hairy if beanstalkd had more include files, because then we'd have to make sure to avoid including them multiple times, and Pike's rule would be a lot more helpful.

But the system include files already do that on their own. As long as for beanstalkd it's just the one file (dat.h), we don't have to worry about protecting dat.h itself from being included multiple times. And compilation seems fast enough.

I guess I feel like that rule isn't actually such a big deal in this particular context. The practical difficulty of using emacs might outweigh any benefit we would get from following that rule in this case.

ysmolsky:
Let me investigate if I can make Emacs handle dat.h better without this change.  For some reason I am not confident in this PR 100% anymore. 

ALL: Please do not merge it yet.

PR #463, opened 2019-07-12
test peek* and kick commands

ysmolsky:
Drain mode was quickly tested as well.

The cttest_peek_bad_format test function is commented
until the issue #464 is fixed.

Updates #424


ysmolsky:
I think there is a bug in beanstalkd on linux because of strtoull. It parses the number that is in fact larger than UINT64_MAX and then tries to look for the job with that ID.

PR #468, opened 2019-07-15
fix flaky deadline-soon test

ysmolsky:
Test should spend more time than ttr to get job released back into the queue.


ysmolsky:
This is to fix broken master. Should be merged before anything else. *Oops* 🌵

ysmolsky:
Thank you, Tom.

PR #481, opened 2019-07-20
fix benchmarks in testsrv.c

### LINKED ISSUE #430: putdelete benchmarks is 20 times slower on Linux
linux https://travis-ci.org/beanstalkd/beanstalkd/jobs/552358996:
```bash
ctbenchheapinsert	 5000000	       786 ns/op
ctbenchheapremove	 1000000	      1347 ns/op
ctbenchmakejob	10000000	       249 ns/op
ctbenchputdelete1k	     100	  39572775 ns/op	   0.03 MB/s
ctbenchputdelete8byte	     100	  39696999 ns/op	   0.00 MB/s
ctbenchputdelete8k	     100	  39573764 ns/op	   0.26 MB/s
```
osx https://travis-ci.org/beanstalkd/beanstalkd/jobs/552358997:
```bash
ctbenchheapinsert	 5000000	       705 ns/op
ctbenchheapremove	 1000000	      1102 ns/op
ctbenchmakejob	 5000000	       384 ns/op
ctbenchputdelete1k	    5000	    709160 ns/op	   1.68 MB/s
ctbenchputdelete8byte	    5000	    494429 ns/op	   0.02 MB/s
ctbenchputdelete8k	    5000	    550173 ns/op	  19.74 MB/s
```
I wonder if it is because clang optimizes by default on higher lever or some problem with the generate code on Linux.

And I think that benchmarking should take more iterations.

/cc @sergeyklay

ysmolsky:
The slowness in benchmarks on Linux was due to long wait time
in the select syscall. Setting socket's TCP_NODELAY helped.

Also see
https://stackoverflow.com/questions/39270419/select-on-socket-slow-in-linux

I added proper resettimers inside of benchmarks to skip the overhead.
This helped to improve the time spent in all of testsrv tests too.

Bechmarking is not ideal yet. It might be capped by the client's limitations.

Fixes #430

### DIFF
--- a/testjobs.c
+++ b/testjobs.c
@@ -132,6 +132,7 @@ ctbench_make_job(int n)
 {
     int i;
     TUBE_ASSIGN(default_tube, make_tube("default"));
+    ctresettimer();
     for (i = 0; i < n; i++) {
         make_job(0, 0, 1, 0, default_tube);
     }
--- a/testserv.c
+++ b/testserv.c
@@ -13,6 +13,7 @@
 #include <netdb.h>
 #include <netinet/in.h>
 #include <netinet/ip.h>
+#include <netinet/tcp.h>
 #include <arpa/inet.h>
 #include <fcntl.h>
 #include <sys/wait.h>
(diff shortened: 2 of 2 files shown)


sergeyklay:
Good catch

kr:
Never mind about my first comment here. I think it would hide actual time spent waiting for the server, which should be counted as time taken for the work.

ysmolsky:
The main fix was about setting the sock flag. The rest contributed only slight overhead. I think It is okay to reset timer in the test because it specifies the overhead part. Technically i could omit resetting the timer but I did for the completeness.

ysmolsky:
Thanks for some additional insights. This helps a lot.

PR #484, opened 2019-07-21
test ms and improve code style of it

ysmolsky:
I have renamed ms.used to ms.len, added comments explaining
the ms structure. Tests were added, including a little confusing
behaviour of function ms_take. The order of taken elements is
not the same as the order of inserts.


ysmolsky:
Awesome! Thanks for spotting it. 🙏

PR #489, opened 2019-07-23
fix the id parsing in all commands and add tests

### LINKED ISSUE #464: more strict parsing of the ID parameter in commands
While testing the `peek` command I have discovered that parsing of the id parameter in commands is not strict enough: 
```
        errno = 0;
        id = strtoull(c->cmd + CMD_PEEKJOB_LEN, &end_buf, 10);
        if (errno) return reply_msg(c, MSG_BAD_FORMAT);
```
Parsing the id should more strict than that, looks at the `read_num` function of an inspiration.

This test should be uncommented after the fix. It should pass on Linux:
```
void
cttest_peek_bad_format()
{
    port = SERVER();
    fd = mustdiallocal(port);
    mustsend(fd, "peek 18446744073709551616\r\n");
    ckresp(fd, "BAD_FORMAT\r\n");

    mustsend(fd, "peek foo111\r\n");
    ckresp(fd, "BAD_FORMAT\r\n");
}
```
Right now it fails with this error:
```
cttest_peek_bad_format: failure
start server port=33365 pid=5240
>38 peek 18446744073709551616
<38 BAD_FORMAT
>38 peek foo111
<38 NOT_FOUND
testserv.c:249: test: strcmp(exp, line) == 0
testserv.c:249: "BAD_FORMAT
" != "NOT_FOUND
"
```
https://travis-ci.org/beanstalkd/beanstalkd/jobs/558005660

ysmolsky:
This change fixes how IDs are parsed in all commands.
They are parsed into uint64 with full spectrum of checks.
I added the read_uint64 function identical to read_uint32 (former read_num).
Also read_uint* functions report values not falling into the range as invalid.

I have renamed definitions *_JOBKICK to *_KICKJOB because it was inverted from
actual command.

Updates #424
Fixes #464

### DIFF
--- a/dat.h
+++ b/dat.h
@@ -71,9 +71,6 @@ typedef int(FAlloc)(int, int);
 // The width is restricted by Jobrec.body_size that is int32.
 #define JOB_DATA_SIZE_LIMIT_MAX 1073741824
 
-// Maximum value (uint32) allowed in pri, delay and ttr parameters
-#define MAX_UINT32 4294967295
-
 // Use this macro to designate unused parameters in functions.
 #define UNUSED_PARAMETER(x) (void)(x)
 
--- a/prot.c
+++ b/prot.c
@@ -34,7 +34,7 @@ size_t job_data_size_limit = JOB_DATA_SIZE_LIMIT_DEFAULT;
 #define CMD_RELEASE "release "
 #define CMD_BURY "bury "
 #define CMD_KICK "kick "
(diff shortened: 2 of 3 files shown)


ysmolsky:
tests discovered a bug on linux. i will embed the fix into this patch...

ysmolsky:
I have repurposed the PR because some bugs were found during more extensive testing. The latest version contains all the tests and the best implementation.

Issue #382, opened 2018-05-21
Memory leak v.10 testheap.c:222


PatriciaSVMonteiro:
In the ```ctbenchheapinsert``` (testheap.c, line 222) a memory allocation is made for the variable j and it is not released at the end of the function, resulting in a memory leak type error. 

```C
void ctbenchheapinsert(int n)
{
    job *j;
    int i;
    j = calloc(n, sizeof *j);
    assert(j);
    for (i = 0; i < n; i++) {
        j[i] = make_job(1, 0, 1, 0, 0);
        assert(j[i]);
        j[i]->r.pri = -j[i]->r.id;
    }
    Heap h = {0};
    h.less = job_pri_less;
    h.rec = job_setheappos;
    ctresettimer();
    for (i = 0; i < n; i++) {
        heapinsert(&h, j[i]);
    }
}
```

JensRantil:
Hi @PatriciaSVMonteiro! Nice find! Would you be interested in submitting a pull request to fix this?

chadchristenson:
I think this is ok.  heapinsert doesn't allocate memory for the object.  It only allocates space on the heap for the pointer.

sjinks:
Actually @PatriciaSVMonteiro is right, there is a leak. This is what valgrind says:

```
ctbenchheapinsert       ==1215== 
==1215== HEAP SUMMARY:
==1215==     in use at exit: 32,208 bytes in 61 blocks
==1215==   total heap usage: 118 allocs, 57 frees, 1,870,928 bytes allocated
==1215== 
==1215== 8 bytes in 1 blocks are definitely lost in loss record 1 of 6
==1215==    at 0x4C31B25: calloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==1215==    by 0x119DAC: ctbenchheapinsert (testheap.c:208)
==1215==    by 0x10BCD0: runbenchn (ct.c:324)
==1215==    by 0x10BF8E: runbench (ct.c:419)
==1215==    by 0x10C253: runallbench (ct.c:471)
==1215==    by 0x10C620: main (ct.c:560)
==1215== 
==1215== 16 bytes in 1 blocks are definitely lost in loss record 2 of 6
==1215==    at 0x4C2FB0F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==1215==    by 0x10EADB: heapinsert (heap.c:84)
[...]

PR #502, opened 2019-07-26
[WIP] Added memcheck to the build matrix

sergeyklay:
Related: #382

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,11 @@ language: c
 addons:
     apt:
       packages:
-          - lcov
+          - valgrind
+
+matrix:
+    allow_failures:
+        - env: MEMCHECK=1
 
 compiler:
     - clang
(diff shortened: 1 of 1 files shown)


sergeyklay:
@ysmolsky I would like to know should we suppress some errors or not

ysmolsky:
> @ysmolsky I would like to know should we suppress some errors or not

I would prefer not to see "still reachable" kind of errors. The most important are "definitely lost" and "possibly lost". We need to pay attention to these first.

http://valgrind.org/docs/manual/faq.html#faq.deflost

.travis.yml
```diff
@@ -77,6 +81,27 @@ jobs:
...
+    - stage: Memcheck
+      os: linux
+      compiler: gcc
+      env: MEMCHECK=1
+      before_script:
+          - export COVERAGE=OFF
+      script:
+          - make $MAKEJOBS
+          - make $MAKEJOBS ct/_ctcheck
+          - export MAKEFLAGS="valgrind --leak-check=full"
```
ysmolsky: We can get rid of it by updating to the last version of "ct". I have fixed it there. Let me create patch.

.travis.yml
```diff
@@ -77,6 +81,27 @@ jobs:
...
+          - |
+            valgrind \
+                --read-var-info=yes \
+                --error-exitcode=1 \
+                --fullpath-after= \
+                --track-origins=yes \
+                --leak-check=full \
+                --show-leak-kinds=all \
```
ysmolsky: Please see my comment about what kind of stuff we do not need to show.

.travis.yml
ysmolsky: PR was created. After the merge we can remove the (kludge) line above.

ysmolsky:
I am not sure what to do with memory leaks. After all we decided not to fix them in the program, which is supposed to be crash-restarted. @sergeyklay any ideas?

sergeyklay:
@ysmolsky I'm sorry for the delay (too busy at work). Can this wait for some (short) time? I'll try to investigate a bit more in the near future.

ysmolsky:
> @ysmolsky I'm sorry for the delay (too busy at work). Can this wait for some (short) time? I'll try to investigate a bit more in the near future.

Sure thing. By all means, no rushing from me.

ysmolsky:
I think, we won't go there yet. I have somewhat reduced interest in this project since autumn has started. I was a little depressed that there is very little support by the community.  What was done is good enough for my work and I will fix only critical bugs. Most likely I will not try to hunt down all kind of memory leaks since this software is intended as crash-only and there is no sense in graceful exit.

sergeyklay:
Ok, let's close this one :)

PR #504, opened 2019-07-27
drop the pointer from the tube typedef

### LINKED ISSUE #458: typedefs of ms, job and tube should not hide the pointer
Make them similar to other typedefs (`Conn`, `Heap`). The code in question:
```
typedef struct ms     *ms;
typedef struct job    *job;
typedef struct tube   *tube;
typedef struct Conn   Conn;
typedef struct Heap   Heap;
typedef struct Jobrec Jobrec;
```

_Maybe, move each typedef next to the corresponding struct definition:_
```
typedef struct Heap   Heap;
struct Heap {
    size_t  cap;
    size_t  len;
    void    **data;
    Less    less;
    Record  rec;
};
```
_This may allow easier navigation to the struct definition from the referencing source code._

ysmolsky:
Fixes #458

### DIFF
--- a/conn.c
+++ b/conn.c
@@ -14,7 +14,7 @@ static uint tot_conn_ct = 0;
 int verbose = 0;
 
 static void
-on_watch(Ms *a, tube t, size_t i)
+on_watch(Ms *a, Tube *t, size_t i)
 {
     UNUSED_PARAMETER(a);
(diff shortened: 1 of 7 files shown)


ysmolsky:
> Finally 👍

Yup, this contraption is gone. But there are some other warts too, hehehe.

Issue #20, opened 2009-10-08
Reprioritise existing jobs


mloughran:
Given a job id, I would like to be able to reprioritize it. Currently I can only achieve this by calling peek, delete, then recreating the job.

One possibility would be to allow release to be called without first having reserved the job. Currently the interface for delete and release is inconsistent:

```
put 0 0 60 3
foo
INSERTED 1
delete 1
DELETED
put 0 0 60 3
bar
INSERTED 2
release 2 100 0
NOT_FOUND
delete 2
DELETED
```

ysmolsky:
Closed in favor of #310

Issue #310, opened 2016-04-03
support reserving a specific job by id


JensRantil:
Idea:

```
reserve-job <id>
```

Output:

```
RESERVED <id> <bytes>\r\n
<data>\r\n
```

or

```
ALREADY_RESERVED\r\n
```

or

```
NOT_FOUND\r\n
```

This would fix #109 and #20 and lots of other cases where one would like to modify a job. The `ALREADY_RESERVED` would make sure that noone is processing the job, before I modify it.

I don't know the code (yet), would it be somewhat efficient to execute `reserve-job <id>`?

JensRantil:
This would also fix #170.

JensRantil:
This would allow #222 to reserve a job, delete it and create it using a different delay.

silviucpp:
I think @kr should approve or not the specs in case he believes will add a benefit to the project. For example I'm ok to spend my time in implementing/fixing different bugs but It will be a waste of time to work on something and to not receive any feedback.

cmer:
This would be great @kr. It would allow the following use case:

- Web page displays a list of 50 "things" to review. Reserve all 50 things for the current user
- User looks at all the things on web page and clicks "approve", which makes a call to the backend
- Backend reserves the jobs again and deletes them.
- If user doesn't approve things fast enough, the jobs go back in the queue.

I'm not sure if this use case is currently possible since the beanstalk connection will likely be different for every request (displaying the web page, and approving the things)

silviucpp:
If @kr approves the specs I can implement this . Is not so complicated

contentfree:
Anyone hear from @kr?

ysmolsky:
This might fix #240 as well by following sequence of commands: 
* `reserve-job id-of-a-job-to-move`
* `put job parameters...`

kr:
This is an *excellent* idea.

One orthogonal new feature that combines with the rest of the system to make several other commands more powerful.

Regarding the semantics, I kind of think it would be good to have this command get the job not just from the ready queue, but from anywhere (delay, buried) as long as the job isn't reserved by another connection. That would let clients also change the delay time (including to 0) of delayed jobs, and to unbury specific jobs.

ysmolsky:
> it would be good to have this command get the job not just from the ready queue, but from anywhere

Yes, this is what I had in my mind too. 

[...]

PR #507, opened 2019-07-28
[Feature] Allow reprioritizing queued jobs

tw99:
Addresses: https://github.com/beanstalkd/beanstalkd/issues/20 which would be a helpful feature to have. This PR will update priority for ready, delayed, and buried jobs - and only sift the jobs tube heap if the job is in a ready state. 

Format: 
```
reprioritize <id> <pri>\r\n
```

Returns:

`REPRIORITIZED\r\n` on success/noop
`NOT_FOUND\r\n` when job id not found
`ALREADY_RESERVED\r\n` when job has already been reserved
~`NOOP\r\n` when updated priority is the same~

Looking forward to any feedback!

### DIFF
--- a/dat.h
+++ b/dat.h
@@ -111,6 +111,8 @@ struct Heap {
 int   heapinsert(Heap *h, void *x);
 void* heapremove(Heap *h, size_t k);
 
+void siftdown(Heap *h, size_t k);
+void siftup(Heap *h, size_t k);
 
 struct Socket {
     int    fd;
--- a/heap.c
+++ b/heap.c
@@ -30,7 +30,7 @@ less(Heap *h, size_t a, size_t b)
 }
 
 
-static void
+void
 siftdown(Heap *h, size_t k)
 {
     for (;;) {
@@ -46,7 +46,7 @@ siftdown(Heap *h, size_t k)
 }
 
 
-static void
+void
 siftup(Heap *h, size_t k)
 {
     for (;;) {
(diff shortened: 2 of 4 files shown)


kr:
Is it important to distinguish between `NOOP` and `REPRIORITIZED` cases? Potentially the noop case could be folded in so there’s just one successful response, and then the command would be idempotent.

tw99:
@kr Thanks for the reply. 

Good point, not necessary to distinguish between `NOOP`/`REPRIORITIZED` cases. Updated the branch

ysmolsky:
@tw99 thanks for the contribution. I know that this is a bit too late to ask, but the discussion of new features does not happen inside of PRs, but on the issue tracker. Right now it is not clear to me that the addition of a command proposed here is the best way to solve #20. For example, there is a possibility of adding a command to reserve by id (#310). That way is more flexible because we can release a job with changed delay too. 

I'm trying my best to get familiar with the code base and to make code conforming to C99 standard. That means to have a code with a least amount of surprises. Also testing and coverage to ensure the correctness even though this program was tested by hundreds of deployments on productions everywhere, that is not a sustainable way. I am not ready for adding new features yet. I am sorry. As I stated somewhere else, I am not in the mode of adding new stuff yet.

tw99:
@ysmolsky No problem, it was a good way to get my feet wet in the codebase regardless - let me know if you'd like me to add more robust tests for this feature or if a decision is made on adding the reserve by ID command. I'd like to attempt a PR in that case

ysmolsky:
I am sorry. I have decided to close it in favor of https://github.com/beanstalkd/beanstalkd/issues/310. I gave it a good time to sink it. That direction is more robust. Let me know if you want to work on #310.

Issue #488, opened 2019-07-22
get rid of named variadic macros that is GNU extension


ysmolsky:
Flag `-pedantic` leads to this error. I do not see how to eliminate this because macro is used  in two forms: `twarn("s")` and `twarn("s %s", a)`. C99 standard does not allow no arguments passed to the variable arguments, also it forbids the use of ## (which removes comma from macro when VA_ARGS is empty). 

It's a reminder, because I don't know if it makes sense to fix this at all.

```
./dat.h:209:24: error: named variadic macros are a GNU extension [-Werror,-Wvariadic-macros]
#define twarn(fmt, args...) warn("%s:%d in %s: " fmt, \
                       ^
./dat.h:211:25: error: named variadic macros are a GNU extension [-Werror,-Wvariadic-macros]
#define twarnx(fmt, args...) warnx("%s:%d in %s: " fmt, \
                        ^
```

kr:
If it has to be fixed, maybe something like this would be palatable:

```c
#define twarnf(fmt, ...) warn("%s:%d in %s: " fmt, \
    __FILE__, __LINE__, __func__, __VA_ARGS__)
#define twarn(s) warn(%s:%d in %s: %s", \
    __FILE__, __LINE__, __func__, s)
```

And similarly for twarnx and twarnxf.

ysmolsky:
Thanks, looks good. And one addition is that difference between `twarn` and `twarnx` is obscure. `twarn` should have been named `twarnerr` and `twarnx` as just `twarn`. I keep stuttering while reading this code. 

Will think more.

kr:
The naming convention comes directly from the `err`, `errx`, `warn`, and `warnx` functions of the std lib. I just added a `t` on the front.

ysmolsky:
Yes. All of those functions require some work. I plan to send a PR soon.

ysmolsky:
First try was not successful. This requires more work and planning. Right now these macros are used for reporting errors in testing and writing errors to stderr, however that it is not capture most likely by parent program, except for docker, nor logging has timestamps attached to it. So this is better be reworked in conjunction with fixing logging... 🤷‍♂

PR #510, opened 2019-07-29
make names of warn* functions and macros consistent

ysmolsky:
twarn* macros and warn* functions names are now consistent.

-err suffix means that the function uses errno to print error message.
no suffix means that the function ignores errno.

Updates #488

### DIFF
--- a/conn.c
+++ b/conn.c
@@ -38,12 +38,12 @@ make_conn(int fd, char start_state, Tube *use, Tube *watch)
     Conn *c;
 
     c = new(Conn);
-    if (!c) return twarn("OOM"), NULL;
+    if (!c) return twarnerr("OOM"), NULL;
 
     ms_init(&c->watch, (ms_event_fn) on_watch, (ms_event_fn) on_ignore);
     if (!ms_append(&c->watch, watch)) {
         free(c);
-        return twarn("OOM"), NULL;
+        return twarnerr("OOM"), NULL;
     }
 
     TUBE_ASSIGN(c->use, use);
--- a/darwin.c
+++ b/darwin.c
@@ -41,7 +41,7 @@ sockinit(void)
 {
     kq = kqueue();
     if (kq == -1) {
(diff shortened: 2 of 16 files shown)


kr:
Hm, it seems to me this PR *introduces* some inconsistency. The names `err`, `errx`, `warn`, and `warnx` are from the C standard library, defined in err.h. Try `man errx`or see https://linux.die.net/man/3/errx.

I think it might be better to stick to the standard naming convention.

In fact, right now I'm not even sure why some of them are reimplemented in util.c, since they have effectively the same behavior as the real ones declared in err.h. Looks like it might be so that tests can override the "program name", using the test name (like `cttest_peek_ok`) instead of the actual program name (always `_ctcheck`).

ysmolsky:
Thanks. So in the first place there was a confusion because existing functions were redefined, and interposition in C should be avoided if possible. And lack of experience with err.h did not help me.

I guess revert is the only option now.

ysmolsky:
I have created a revert. I really do not like how C has chosen names for warn* functions. But if we have to stick with them, so be it.

ysmolsky:
We still can use macroses but with old naming scheme:
with error:
```
warn() 
twarn() 
twarnf()
```
Without error:
```
warnx()
twarnx() 
twarnxf()
```

kr:
> So in the first place there was a confusion because existing functions were redefined

At the very least I should have added comments about this in the first place.

> I really do not like how C has chosen names for warn* functions. But if we have to stick with them, so be it.

I don't think we absolutely have to stick with them. My phrasing above was probably too strong. Mostly I wanted to make sure we're fully aware of the existing conventions before changing them, so both sides of the tradeoff can be weighed.

At this point I think it should be your call. You will have to work more than I will with whatever names we pick.

ysmolsky:
Thanks Keith, that is very kind of  you. I will take a pause on this and think about it.

PR #512, opened 2019-07-29
add macros twarnf and twarnerrf for variable arguments

### LINKED ISSUE #488: get rid of named variadic macros that is GNU extension

ysmolsky:
This patch uses __VA_ARGS__ instead of GNU extentsion.
If there is a need to supply formatting string with argument to twarn*
then functions with suffix "f" must be used.

Fixes #488

### DIFF
--- a/conn.c
+++ b/conn.c
@@ -38,12 +38,16 @@ make_conn(int fd, char start_state, Tube *use, Tube *watch)
     Conn *c;
 
     c = new(Conn);
-    if (!c) return twarnerr("OOM"), NULL;
+    if (!c) {
+        twarnerr("OOM");
+        return NULL;
+    }
 
     ms_init(&c->watch, (ms_event_fn) on_watch, (ms_event_fn) on_ignore);
     if (!ms_append(&c->watch, watch)) {
         free(c);
-        return twarnerr("OOM"), NULL;
+        twarnerr("OOM");
+        return NULL;
     }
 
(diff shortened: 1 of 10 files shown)


kr:
I agree it's unfortunate these names don't have an `f` at the end, but the non-f names are the standard names. (See my other comment https://github.com/beanstalkd/beanstalkd/pull/510#issuecomment-516029117).

kr:
Heh, I remember I might have suggested using the `f` suffix in the first place, to help get rid of the GNU preprocessor extensions. Not sure what to suggest. :/

ysmolsky:
This is too ugly and pleasing the compiler this way is too much.

PR #514, opened 2019-07-29
fix ternary operators to not use GNU extension

ysmolsky:
Discovered by -std=c99 -pedantic

### DIFF
--- a/prot.c
+++ b/prot.c
@@ -1364,11 +1364,16 @@ dispatch_cmd(Conn *c)
             return reply_msg(c, MSG_BAD_FORMAT);
         op_ct[type]++;
 
-        j = job_find(id);
-        j = remove_reserved_job(c, j) ? :
-            remove_ready_job(j) ? :
-            remove_buried_job(j) ? :
-            remove_delayed_job(j);
+        {
+            Job *jf = job_find(id);
+            j = remove_reserved_job(c, jf);
+            if (!j)
+                j = remove_ready_job(jf);
+            if (!j)
+                j = remove_buried_job(jf);
+            if (!j)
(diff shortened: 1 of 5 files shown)


prot.c
```diff
@@ -1365,10 +1365,13 @@ dispatch_cmd(Conn *c)
...
         j = job_find(id);
-        j = remove_reserved_job(c, j) ? :
-            remove_ready_job(j) ? :
-            remove_buried_job(j) ? :
-            remove_delayed_job(j);
+        Job *jr;
+        if ((jr = remove_reserved_job(c, j)) ||
+            (jr = remove_ready_job(j)) ||
+            (jr = remove_buried_job(j)) ||
+            (jr = remove_delayed_job(j))) {
```
kr: We want `j` to be NULL if none of the four cases are found. But this looks to me like it will not do that: if `j` is a non-NULL job and none of these four cases succeed, then the body of the if statement will not execute and `j` will continue to be a non-NULL job.

kr: Maybe this would work:

```suggestion
Job *jr = job_find(id);
(j = remove_reserved_job(c, jr)) ||
    (j = remove_ready_job(jr)) ||
    (j = remove_buried_job(jr)) ||
    (j = remove_delayed_job(jr));
```

ysmolsky: You are right. Let me check why it was not caught by tests.

ysmolsky: There was not test for this case. I have added it and fixed the bug.

ysmolsky: And I had to not use || contraption because of unused complain from the compiler.

ysmolsky: Also I have put this into { } block to isolate the scope of var `jf`.

PR #515, opened 2019-07-30
do not return void expressions

ysmolsky:
It is confusing to understand and it is against C99.

Also remove redundant returns in end of of functions.


ysmolsky:
The change is big, but changes were rather mechanical in all of cases.

Issue #442, opened 2019-07-05
all: reduce complex lines in the code


ysmolsky:
Don't get me wrong. I like the most of the style adopted in beanstalkd. There are some things that can be improved:

1. `return`'s on the same line as `if`. They are common. `return` should start a new line always.
```
    if (!c->out_job) return reply_serr(c, MSG_OUT_OF_MEMORY);
[skip]
    if (r > stats_len) return reply_serr(c, MSG_INTERNAL_ERROR);
```
2. Comma expressions in returns. Comma expressions. Not so common. Do not use comma expression where multiple lines can do.
```
if (some) twarnx("..."), return expression;
```
I prefer to read the code vertically, but this style makes by eyes bleed because I have to read the code in double columns. I hope that you agree with me and we can fix the code little by little, moving forward, or at least not add mode code in this style.

I prefer verbose clarity rather than obscure smartness.

@tomponline @sergeyklay

kr:
For whatever it's worth, I 100% agree with you. I think I was wrong to use that style.

ysmolsky:
The main bulk of it was done, but the remains are to be fixed as we go. Closing

PR #522, opened 2019-08-04
fix style of the code

ysmolsky:
* remove global vars where possible
* reduce scope of variables
* remove shadowing of variables
* remove redundant assigments
* fix prototypes according to the definion of functions

Updates #442

### DIFF
--- a/file.c
+++ b/file.c
@@ -338,7 +338,6 @@ readrec5(File *f, Job *l, int *err)
             j = make_job_with_id(jr.pri, jr.delay, jr.ttr, jr.body_size,
                                  t, jr.id);
             j->next = j->prev = j;
-            j->r.created_at = jr.created_at;
         }
         j->r.id = jr.id;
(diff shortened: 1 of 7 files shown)


file.c
```diff
@@ -338,7 +338,6 @@ readrec5(File *f, Job *l, int *err)
             j = make_job_with_id(jr.pri, jr.delay, jr.ttr, jr.body_size,
                                  t, jr.id);
             j->next = j->prev = j;
-            j->r.created_at = jr.created_at;
```
kr: Is this line really not necessary?

ysmolsky: Yep, most of the issues here were reported by cppcheck and manually fixed by me.
In this case, see line 348 where value to `j->r.created_at` assigned regardless of deleted line the branch `if (!j)`.

kr: Oh yes, indeed!

PR #523, opened 2019-08-04
ms.c: simplify the code checking OOM in the grow function

ysmolsky:
Updates #442


ysmolsky:
Right, so travis got broken.

PR #525, opened 2019-08-04
refactor the code controlling events for sockets

### LINKED ISSUE #524: prot.c: what is the purpose of dirty
I cannot quite grasp the meaning and idea behind the dirty variable. Especially `protrmdirty()` function drives me wild. I just do not understand what for is that?

@kr, please drop some hints.

This is exactly all the places where it's being used.
```c
static Conn *dirty;

static void
reply(Conn *c, char *line, int len, int state)
{
    if (!c) return;

    connwant(c, 'w');
    c->next = dirty;
    dirty = c;
    c->reply = line;
    c->reply_len = len;
    c->reply_sent = 0;
    c->state = state;
    if (verbose >= 2) {
        printf(">%d reply %.*s\n", c->sock.fd, len-2, line);
    }
}


static void
protrmdirty(Conn *c)
{
    Conn *x, *newdirty = NULL;

    while (dirty) {
        x = dirty;
        dirty = dirty->next;
        x->next = NULL;

        if (x != c) {
            x->next = newdirty;
            newdirty = x;
        }
    }
    dirty = newdirty;
}

static void
wait_for_job(Conn *c, int timeout)
{
    c->state = STATE_WAIT;
    enqueue_waiting_conn(c);

    /* Set the pending timeout to the requested timeout amount */
    c->pending_timeout = timeout;

    connwant(c, 'h'); // only care if they hang up
    c->next = dirty;
    dirty = c;
}

static void
reset_conn(Conn *c)
{
    connwant(c, 'r');
    c->next = dirty;
    dirty = c;

    /* was this a peek or stats command? */
    if (c->out_job && c->out_job->r.state == Copy) job_free(c->out_job);
    c->out_job = NULL;

    c->reply_sent = 0; /* now that we're done, reset this */
    c->state = STATE_WANTCOMMAND;
}

static void
update_conns()
{
    int r;
    Conn *c;

    while (dirty) {
        c = dirty;
        dirty = dirty->next;
        c->next = NULL;
        r = sockwant(&c->sock, c->rw);
        if (r == -1) {
            twarn("sockwant");
            connclose(c);
        }
    }
}

static void
h_conn(const int fd, const short which, Conn *c)
{
    if (fd != c->sock.fd) {
        twarnx("Argh! event fd doesn't match conn fd.");
        close(fd);
        connclose(c);
        update_conns();
        return;
    }

    if (which == 'h') {
        c->halfclosed = 1;
    }

    conn_data(c);
    while (cmd_data_ready(c) && (c->cmd_len = cmd_len(c))) {
        dispatch_cmd(c);
        fill_extra_data(c);
    }
    if (c->state == STATE_CLOSE) {
        protrmdirty(c);
        connclose(c);
    }
    update_conns();
}

```

ysmolsky:
The name "dirty" did not reveal the purpose of the variable.
I've renamed it to "epollq".

I have removed connwant() and moved its functionality into introduced
function epollq_add(). This function adds a connection as pending for
polling change. protrmdirty() was renamed to epollq_rmconn().
update_conns() was renamed to epollq_apply().

I added a test checking the "quit" command.

Fixes #524

### DIFF
--- a/conn.c
+++ b/conn.c
@@ -138,16 +138,6 @@ conntickat(Conn *c)
 }
 
 
-// TODO: remove this function by inlining its content into 3 callees places.
-// Reason: conn.c does not use rw anywhere in this file.
-void
-connwant(Conn *c, int rw)
-{
-    c->rw = rw;
-    connsched(c);
-}
-
-
 // Remove c from the c->srv heap and reschedule it using the value
 // returned by conntickat if there is an outstanding timeout in the c.
 void
--- a/dat.h
+++ b/dat.h
@@ -368,7 +368,6 @@ struct Conn {
 };
 int  conn_less(void *ca, void *cb);
 void conn_setpos(void *c, size_t i);
-void connwant(Conn *c, int rw);
 void connsched(Conn *c);
 void connclose(Conn *c);
 void connsetproducer(Conn *c);
--- a/prot.c
+++ b/prot.c
@@ -235,6 +235,9 @@ static char instance_hex[instance_id_bytes * 2 + 1]; // hex-encoded len of insta
 static struct utsname node_info;
 static uint64 op_ct[TOTAL_OPS], timeout_ct = 0;
 
+// Single linked list that holds connections
+// that required update in polling mechanism by update_conns().
+// TODO: rename to epoll_changes?
 static Conn *dirty;
 
 static const char * op_names[] = {
@@ -274,33 +277,26 @@ buried_job_p(Tube *t)
 }
 
 static void
-reply(Conn *c, char *line, int len, int state)
-{
-    if (!c) return;
-
-    connwant(c, 'w');
+dirty_add(Conn *c, char rw) {
+    c->rw = rw;
+    connsched(c);
     c->next = dirty;
     dirty = c;
-    c->reply = line;
-    c->reply_len = len;
-    c->reply_sent = 0;
-    c->state = state;
-    if (verbose >= 2) {
-        printf(">%d reply %.*s\n", c->sock.fd, len-2, line);
-    }
 }
 
-
+// Remove connection c from the dirty.
 static void
-protrmdirty(Conn *c)
+dirty_rmconn(Conn *c)
 {
     Conn *x, *newdirty = NULL;
 
     while (dirty) {
+        // x as next element from dirty.
         x = dirty;
         dirty = dirty->next;
         x->next = NULL;
 
+        // put x back into newdirty list.
         if (x != c) {
             x->next = newdirty;
             newdirty = x;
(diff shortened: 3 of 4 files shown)


ysmolsky:
I have went further and renamed all the wanted functions and removed TODO about it. This change looks finished to me now. PTAL.

ysmolsky:
Please take a look

conn.c
```diff
@@ -138,16 +138,6 @@ conntickat(Conn *c)
 }
 
 
-// TODO: remove this function by inlining its content into 3 callees places.
-// Reason: conn.c does not use rw anywhere in this file.
-void
-connwant(Conn *c, int rw)
```
kr: I was trying to factor some things out of prot.c so the struct definition `struct Conn { ... }` could be moved to conn.c (and `typedef struct Conn Conn` would of course stay in dat.h). That's why this function exists, it was an early step in that refactoring. But of course I didn't get far enough to actually move the struct definition.

I was even hoping to do that for most of the struct definitions some day.

ysmolsky: Aha, this is why so many `conn*` functions, and some of them are somewhat forced... This does not contradict to my intentions. I have isolated operations on `epollq` (former `dirty`) into functions that can/should be moved into `conn.c`. The same is true about `epollq` that can be hidden in `conn.c` I want to break some global dependencies, but it's hard when everything depends on everything. My decision about this might be not ideal and I can change it later but for now it looks reasonable.

ysmolsky: Q about moving struct def of `Conn` out of `data.h`? Do you think it's possible because clients would still have to see `*Conn` to be able to interface with functions in `conn.c`?

kr: Yeah, I agree you should approach refactoring in whatever way best fits your longer term goals.

It's possible to move the struct declaration to conn.c if client functions use only `*Conn` and never `Conn`. Here's a small example:

ex.h:
```c
typedef struct Blob Blob;
Blob *blob_new();
void blob_test(Blob *b);
```

blob.c:
```c
#include <stdlib.h>
#include <stdio.h>
#include "ex.h"

struct Blob {
    int n;
};

Blob *
blob_new()
{
	Blob *b = malloc(sizeof(Blob));
	b->n = 2;
	return b;
}

void
blob_test(Blob *b)
{
	printf("%d\n", b->n);
}
```

user.c:
```c
#include "ex.h"

int
main()
{
	Blob *b = blob_new();
	blob_test(b);
}
```

You can compile this with `cc *.c` and running `./a.out` should print `2`.

But things like like `malloc(sizeof(Blob))` in user.c won't work.

ysmolsky: Thanks, it looks like data encapsulation to me. I never knew about this in C.

ysmolsky: @kr I have a strange question about former "dirty" list. Why do we have that in a first place? Why cannot we call sockwant() exactly in place without putting the conn to delayed list "dirty" first? I cannot find explanation for this, understanding this mechanism feels important.

Issue #477, opened 2019-07-18
fsync usage needs audit


kr:
Fsync, as implemented on Linux, has some semantics the Postgres devs found extremely surprising. Email thread: <https://danluu.com/fsyncgate/>

When I read that email thread I found it surprising, too. This might very well have implications for beanstalkd, since it means I originally wrote beanstalkd assuming an incorrect set of semantics for fsync.

Ideally, someone should take the time to understand fsync's true behavior, and audit beanstalkd's usage of fsync to make sure that we can provide the guarantees we claim to provide, taking into consideration fsync's true guarantees.

We are probably in better shape than postgres, since beanstalkd already uses the "single-process, single-file-descriptor write() then fsync()" approach. ([Search for the message](https://danluu.com/fsyncgate/) with `Date:2018-04-04 06:00:21`.) We would need to add the "on failure, retry all work since last successful fsync()" part, and that is still not easy.

A quick look at the code in walg.c shows our existing usage of fsync is much too loose. On error, all it does is print a warning to the log and continue as if fsync had succeeded.

ysmolsky:
What happens if beanstalkd is killed with -TERM in the middle of writing files (write() then fsync())? I guess the same as when it is killed with -KILL

ysmolsky:
It could be slightly related to #496.

ysmolsky:
Origin of the fsyncgate link: https://danluu.com/deconstruct-files/. Good read.

ysmolsky:
Good summary from PG community on fsync() issue: https://wiki.postgresql.org/wiki/Fsync_Errors.

pgsql PANICs on fsync error.

MacOS does not write flush unless `F_FULLFSYNC` is used.  man page:
>   For applications that require tighter guarantees about the integrity of their data, Mac OS X
 provides the F_FULLFSYNC fcntl.  The F_FULLFSYNC fcntl asks the drive to flush all buffered
 data to permanent storage.  Applications, such as databases, that require a strict ordering of
 writes should use F_FULLFSYNC to ensure that their data is written in the order they expect.
  Please see fcntl(2) for more detail.

ysmolsky:
Initial audio is not reassuring:
* Errors from fsync() are just ignored
* Errors in walmaint() are also ignored. 
* Parent directory is never fsynced, and that is cruical to ensure that file was updated in full
* IO errors are not properly reported to the client, #535 
* What should beanstalkd do when write() or fsync() has failed not because of full disk?

[...]

PR #528, opened 2019-08-06
handle SIGTERM to allow quick exit under Docker

### LINKED ISSUE #527: beanstalkd does not exit on -TERM signal in Docker
Something is off when beanstalkd is ran in docker container as first process.
```
% docker run -it --pid=container:bean --net=container:bean --cap-add sys_admin alpine sh
/ # kill -TERM 1
/ # kill -TERM 1
/ # kill -KILL 1
/ # kill -INT 1
/ # kill -HUP 1
/ #
```
I have attached strace. Notice that nothing happened for -KILL signal:
```
% docker run -t --pid=container:bean --net=container:bean --cap-add sys_admin --cap-add sys_ptrace strace
strace: Process 1 attached
epoll_pwait(6, 0x7ffff54d60dc, 1, 3600000, NULL, 8) = -1 EINTR (Interrupted system call)
--- SIGTERM {si_signo=SIGTERM, si_code=SI_USER, si_pid=14, si_uid=0} ---
epoll_pwait(6, 0x7ffff54d60dc, 1, 3600000, NULL, 8) = -1 EINTR (Interrupted system call)
--- SIGTERM {si_signo=SIGTERM, si_code=SI_USER, si_pid=14, si_uid=0} ---
epoll_pwait(6, 0x7ffff54d60dc, 1, 3600000, NULL, 8) = -1 EINTR (Interrupted system call)
--- SIGINT {si_signo=SIGINT, si_code=SI_USER, si_pid=14, si_uid=0} ---
epoll_pwait(6, 0x7ffff54d60dc, 1, 3600000, NULL, 8) = -1 EINTR (Interrupted system call)
--- SIGHUP {si_signo=SIGHUP, si_code=SI_USER, si_pid=14, si_uid=0} ---
epoll_pwait(6,
```

Dockerfile that uses beanstalkd 1.10 build on alpine:
```
FROM alpine
RUN apk add --no-cache beanstalkd
RUN mkdir /beanstalkd
EXPOSE 11300
ENTRYPOINT ["/usr/bin/beanstalkd", "-z", "650000000"]
CMD ["-b", "/beanstalkd", "-f", "300"]
```
I have tried debian:jessie with the same result. So it is not Alpine linux.

The code where it happens:
```
    r = epoll_wait(epfd, &ev, 1, (int)(timeout/1000000));
    if (r == -1 && errno != EINTR) {
        twarn("epoll_wait");
        exit(1);
    }
```
It means that on regular clean Linux when I sent -TERM to the process return from epoll_wait never happens and process is immediately cleaned by OS. But on Docker the call returns and r==-1, errno=EINTR.

ysmolsky:
Before this patch: when beanstalkd with PID=1 received SIGTERM
running inside of Docker container, the process never exited.
Docker SIGKILLed the process after 10s timeout.

This change adds an ability to handle such case without timeout.
Server will signal itself SIGKILL when SIGTERM is received.
This allows instant termination. This is part of the crash-only
strategy so crashing this way is okay.

Fixes #527

### DIFF
--- a/prot.c
+++ b/prot.c
@@ -14,6 +14,7 @@
 #include <netinet/in.h>
 #include <inttypes.h>
 #include <stdarg.h>
+#include <signal.h>
 
 /* job body cannot be greater than this many bytes long */
 size_t job_data_size_limit = JOB_DATA_SIZE_LIMIT_DEFAULT;
@@ -226,7 +227,10 @@ static struct stats global_stat = {0, 0, 0, 0, 0, 0, 0};
 
 static Tube *default_tube;
 
-static int drain_mode = 0;
+// If drain_mode is 1, then server does not accept new jobs.
+// Variable is set by the SIGUSR1 handler.
+static volatile sig_atomic_t drain_mode = 0;
+
 static int64 started_at;
 
 enum { instance_id_bytes = 8 };
@@ -275,6 +279,9 @@ buried_job_p(Tube *t)
     return job_list_any_p(&t->buried);
 }
 
+// epollq_add schedules connection c in the s->conns heap, adds c
+// to the epollq list to change expected operation in event notifications.
+// rw='w' means to notify when socket is writeable, 'r' - readable, 'h' - closed.
 static void
 epollq_add(Conn *c, char rw) {
     c->rw = rw;
@@ -283,7 +290,7 @@ epollq_add(Conn *c, char rw) {
     epollq = c;
 }
 
-// Remove connection c from the epollq.
+// epollq_rmconn removes connection c from the epollq.
 static void
 epollq_rmconn(Conn *c)
 {
@@ -304,8 +311,8 @@ epollq_rmconn(Conn *c)
     epollq = newhead;
 }
 
-// Propagate changes to event notification mechanism about wanted responses
-// from connections. Clear the epollq list.
+// Propagate changes to event notification mechanism about expected operations
+// in connections' sockets. Clear the epollq list.
 static void
 epollq_apply()
 {
--- a/main.c
+++ b/main.c
@@ -37,6 +37,12 @@ su(const char *user)
 }
 
 static void
+handle_sigterm_pid1()
+{
+    raise(SIGKILL);
+}
+
+static void
 set_sig_handlers()
 {
     struct sigaction sa;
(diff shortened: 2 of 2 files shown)


kr:
Beanstalkd was designed as a [crash-only](https://www.usenix.org/conference/hotos-ix/crash-only-software) system. It should not be necessary to do a graceful shutdown.

It's possible that we could have [bugs](https://github.com/beanstalkd/beanstalkd/issues/477) that make this not work as designed, but I think in that case a better solution would be to fix any bugs so that crashes are safe.

Regarding beanstalkd not exiting inside docker, I think a better solution would be to avoid running beanstalkd as pid 1. If that's not possible, then a simpler version of this patch would be to handle SIGTERM by converting it to SIGKILL (roughly by doing `kill(getpid(), SIGKILL)` in the SIGTERM handler).

ysmolsky:
Keith, thank you for reviewing this. 

Is there anything to read about the promises of beanstalkd regarding crashes? Source code does not say what is expected of it.

The only place in the doc that I could find:
> * `-f` <ms>:
  Call fsync(2) at most once every <ms> milliseconds. Larger values
  for <ms> reduce disk activity and improve speed at the cost of
  safety. A power failure could result in the loss of up to <ms>
  milliseconds of history.

ysmolsky:
I have learned about this in sources and in the paper. I see how it works now. I will repurpose this issue to fix it for Docker with PID=1 case only. It does not make sense to ask users to handle this situation separately by running some wrapper around beanstalkd, so the server pid will be more than 1. The fix is trivial and it handles only the case of Docker since elsewhere it won't be run with pid=1.

ysmolsky:
I have repurposed the PR.

kr:
Thanks for taking the trouble to figure this all out! (And sorry for not having documentation for any of this haha.)

prot.c
```diff
@@ -276,6 +280,9 @@ buried_job_p(Tube *t)
     return job_list_any_p(&t->buried);
 }
 
+// Put the next expected operation 'rw' on the c connection to the epollq list.
+// rw='w' means to notify when socker is writeable, 'r' - readable,
```
kr: s/socker/socket/

ysmolsky:
> Thanks for taking the trouble to figure this all out! (And sorry for not having documentation for any of this haha.)

All this is like HUGE puzzle, better than any puzzle game so far. 😉  I learn a lot. For instance I had no idea that data from write() -> buffers in kernel are flushed when process is SIGKILLed, and it is not lost. Only stdio buffers are flushed when process exit() normally so exit handles are triggered. 

I have just started reading that talk from Danluu and I am astonished at complexity of handling files properly. I think beanstalkd does not do it right at all. It tries, but not close enough. I just wish that there was someone who could share this awesomeness and had the same kind of excitement for the project. 🤣 

I might need to introduce a way to test a failing fsync, but I have no idea how to replicate that on low level. Mocking up fsync is easy but does not reflect how things behave in kernel.

kr:
> I might need to introduce a way to test a failing fsync, but I have no idea how to replicate that on low level. Mocking up fsync is easy but does not reflect how things behave in kernel.

The linux kernel has some fault-injection abilities. https://www.kernel.org/doc/html/latest/fault-injection/fault-injection.html

That combined with qemu or user-mode-linux or something might be good for testing this sort of thing. But that's just a thought, I have never really investigated this in depth.

PR #529, opened 2019-08-07
add comments about Conn and Socket structs

ysmolsky:


tomponline:
@ysmolsky always useful to have some comments :)

ysmolsky:
> @ysmolsky always useful to have some comments :)

Thanks! Even though, some of them might be a little off, they are better than nothing, so others dont have to untangle the puzzle. I hope.

PR #531, opened 2019-08-07
linux.c: fix rawfalloc to return errno on error

ysmolsky:
ftruncate() does not return errno.

### DIFF
--- a/linux.c
+++ b/linux.c
@@ -22,7 +22,15 @@ static int epfd;
 int
 rawfalloc(int fd, int len)
 {
-    return ftruncate(fd, len);
+    // XSI-conformant systems might not extend the file and return an error.
+    // ftruncate() might extend the file with a sequence of null bytes or a hole.
+    // Latter means that disk blocks are not allocated before the write happens.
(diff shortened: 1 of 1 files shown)


ysmolsky:
@kr PTAL, it might be related to the promises that walwrite() makes:
```
// Walwrite writes j to the log w (if w is enabled).
// On failure, walwrite disables w and returns 0; on success, it returns 1.
// Unlke walresv*, walwrite should never fail because of a full disk.
// If w is disabled, then walwrite takes no action and returns 1.
int
walwrite(Wal *w, Job *j)
```
If rawfalloc has created a file with holes, then walwrite cannot guarantee that disk full won't happen.

kr:
> If rawfalloc has created a file with holes, then walwrite cannot guarantee that disk full won't happen.

Ugh, I see. This function formerly used `posix_fallocate` (as you may have seen) but it got changed in #341 because `posix_fallocate` isn't available everywhere and I didn't verify that `ftruncate` actually allocates disk blocks.

I probably should have just declined to merge #341 — correctness is more important than portability. Arguably it should be changed back before the next release.

PR #533, opened 2019-08-08
Free placeholder stats/peek job by checking state in connclose

tw99:
"Placeholder" jobs used to return info (stats, list tube) have their state set to `Copy`, as well as copied jobs for peek commands. Consistent with https://github.com/beanstalkd/beanstalkd/blob/master/prot.c#L1859

There could also be a potential memory leak if a connection is closed with a pending peek job copy (where `c->out_job->r.id` is set).

Sidenote: should there be another enum constant added to better indicate the state of these fake jobs - rather than `Copy`?

### DIFF
--- a/conn.c
+++ b/conn.c
@@ -226,7 +226,7 @@ connclose(Conn *c)
     job_free(c->in_job);
 
     /* was this a peek or stats command? */
-    if (c->out_job && !c->out_job->r.id) job_free(c->out_job);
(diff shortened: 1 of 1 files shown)


ysmolsky:
That's a good one. I have traced that code path and I see that it can lead to memory leak in case of closed conn right after issuing the peek-job command.

Adding new enum constant should be motivated if Fake is different from Copy . Copy and Fake smell the same to me right now.

ysmolsky:
Thank you very much for the help!

PR #541, opened 2019-08-11
factor out the code acquiring the wal dir

ysmolsky:
This removed duplication of code between the testing and server sites.
Also it enables adding specialized function for forking the server in
tests, namely testing of AF_UNIX in #150.

### DIFF
--- a/dat.h
+++ b/dat.h
@@ -478,5 +478,6 @@ struct Server {
     // Connections that must produce deadline or timeout, ordered by the time.
     Heap   conns;
 };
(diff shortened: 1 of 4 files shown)


ysmolsky:
@etanol, I will help you with testing of your PR #150, and I did this change. Details will follow. I will think more about testing of AF_UNIX and let you know in the comments.

We need to merge this patch before yours.

etanol:
Ok, no problem.  Go ahead.

PR #543, opened 2019-08-13
rename delay_q_peek() to soonest_delayed_job()

ysmolsky:
I have removed the delay_q_take() function because it had
only one line added compared to delay_q_peek(). That line
was put in place. As a result prottick() became easier to read.


ysmolsky:
Thanks.

I do not know maybe it's only me... Not sure why this code lays in such ruins. I hope I am not overdoing this.

PR #544, opened 2019-08-13
disable forced binding to IPV6-only stack if "-l" was not set

ysmolsky:
This issue appears on FreeBSD 8.4 where IPV6_V6ONLY is set to 1 by
default. This patch unsets this flag if host was not specified.

Each OS choses the default value for this flag, but it's better
to override it for the case when listening on all interface is
expected.

Fixes #228

### DIFF
--- a/net.c
+++ b/net.c
@@ -38,7 +38,7 @@ make_inet_socket(char *host, char *port)
     struct addrinfo *airoot, *ai, hints;
 
     memset(&hints, 0, sizeof(hints));
-    hints.ai_family = PF_UNSPEC;
+    hints.ai_family = AF_UNSPEC;
     hints.ai_socktype = SOCK_STREAM;
     hints.ai_flags = AI_PASSIVE;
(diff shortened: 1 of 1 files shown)


net.c
```diff
@@ -86,6 +86,16 @@ make_inet_socket(char *host, char *port)
             continue;
         }
 
+        if (host == NULL && ai->ai_family == AF_INET6) {
+            flags = 0;
+            r = setsockopt(fd, IPPROTO_IPV6, IPV6_V6ONLY, &flags, sizeof(flags));
+            if (r == -1) {
+                twarn("setting IPV6_V6ONLY off on fd %d", fd);
```
tomponline: @ysmolsky can you change this warning to "setting IPV6_V6ONLY failed ..." rather than "off", as it suggests beanstalkd has decided to disable it rather than just failing to enable it due to some external issue. Thanks

ysmolsky: Done, and I have corrected the commit msg and descr.

PR #547, opened 2019-08-15
rename deadline_at to unpause_at

ysmolsky:
And do collateral renames to improve clarity.
Add a check for not NULL invariant.


ysmolsky:
Thank you, Tom.

PR #551, opened 2019-08-17
improve clarity for job_list_* operations

ysmolsky:
Bunch of functions renamed here and there.
remote_waiting_conn() does not return anything anymore,
instead it is noop for not waiting connection.

reserve_job() body mostly was moved to conn_reserve_job().
the rest was left in place of the call to it.


ysmolsky:
Thanks

PR #557, opened 2019-08-23
add the OS version and platform to the stats command

ysmolsky:
The motivation is to have more detailed stats command so that users can report it when they have some problems.

Make counters as uint64 where possible, and rearrange the code.


ysmolsky:
Please take a look.

PR #558, opened 2019-08-27
remove the -c and -n flags that control log compaction

### LINKED ISSUE #552: remove -n option as non functional and redundant
There is no point to keep that option.

> -n option was only useful during development of the
compaction feature to compare its performance and disk
use to the older non-compacting behavior.
Which reminds me, we should remove that option now.
There's no reason for anyone to use it.
_Originally posted by @kr in https://github.com/beanstalkd/beanstalkd/issues/15#issuecomment-11368695_

ysmolsky:
Log compaction should not be switched off.

Fixes #552

### DIFF
--- a/doc/beanstalkd.ronn
+++ b/doc/beanstalkd.ronn
@@ -28,13 +28,6 @@ and format of the `beanstalkd` protocol.
   in <path>, then, during normal operation, append new jobs and
   changes in state to the binlog.
 
-* `-c`:
-  Perform online, incremental compaction of binlog files. Negates
-  `-n`. This is the default behavior.
-
-  (Do not use this option, except to negate `-n`. Both `-c` and `-n`
-  will likely be removed in a future `beanstalkd` release.)
-
 * `-f` <ms>:
   Call fsync(2) at most once every <ms> milliseconds. Larger values
   for <ms> reduce disk activity and improve speed at the cost of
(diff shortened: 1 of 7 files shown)


util.c
```diff
@@ -109,8 +109,8 @@ usage(int code)
             " -z BYTES set the maximum job size in bytes (default is %d, max allowed is %d)\n"
             " -s BYTES set the size of each write-ahead log file (default is %d)\n"
             "            (will be rounded up to a multiple of 4096 bytes)\n"
-            " -c       compact the binlog (default)\n"
-            " -n       do not compact the binlog\n"
+            " -c       (DEPRECATED) compact the binlog (default)\n"
+            " -n       (DEPRECATED) do not compact the binlog\n"
```
sergeyklay: The "deprecated" term means "still works but not recommended", but the behavior was actually changed, isn't?

ysmolsky: Oh right, bad choice of words. In fact it was completely removed, not only discouraged.

ysmolsky:
I have removed these flags from the help because they should not be used. In case if somebody was using them, he will get a warning.

ysmolsky:
@kr thank you. I have fixed manuals.

doc/beanstalkd.ronn
```diff
@@ -96,6 +83,17 @@ and format of the `beanstalkd` protocol.
 * `-z` <bytes>:
   The maximum size in bytes of a job.
 
+* `-c`:
+  This flag has no effect. It is kept for historical compatibility only.
+
+  Perform online, incremental compaction of binlog files. Negates
```
kr: Do you think this line should be deleted?

ysmolsky: They have no meaning. Removed them.

PR #562, opened 2019-09-06
fix response for the command that is too long

### LINKED ISSUE #337: a command without parameters should not yield `UNKNOWN_COMMAND`
```
put
UNKNOWN_COMMAND
```

If I execute `put` without arguments, I get `UNKNOWN_COMMAND`. Since `put` is a valid command, but I am missing parameters, I am expecting `BAD_FORMAT`. As soon as I add a single parameter I get `BAD_FORMAT`:
```
put 0
BAD_FORMAT
```

ysmolsky:
Before the patch, if a client sent too long command, server would
respond by BAD_FORMAT, followed by UNKNOWN_COMMAND. This resulted
in desync between client and server.

This patch adds another state STATE_WANT_ENDLINE,
whose purpose is to drop command line until the EOL is found.
Client can send as big command as he want and server will be able
to skip it and return only one error message: BAD_FORMAT.

Some STATE_* constants were renamed to improve readability.

Fixes #337

### DIFF
--- a/doc/protocol.txt
+++ b/doc/protocol.txt
@@ -42,7 +42,8 @@ the server will reply with one of the following error messages:
    http://groups.google.com/group/beanstalk-talk.
 
  - "BAD_FORMAT\r\n" The client sent a command line that was not well-formed.
-   This can happen if the line does not end with \r\n, if non-numeric
+   This can happen if the line's length exceeds 224 bytes including \r\n,
+   if the name of a tube exceeds 200 bytes, if non-numeric
    characters occur where an integer is expected, if the wrong number of
    arguments are present, or if the command line is mal-formed in any other
    way.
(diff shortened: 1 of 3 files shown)


sergeyklay:
@ysmolsky Should we update docs as well?

ysmolsky:
> @ysmolsky Should we update docs as well?

Docs mention that already, but expanding it a little would not hurt I guess.
https://github.com/beanstalkd/beanstalkd/blob/master/doc/protocol.txt#L44

ysmolsky:
I have fixed the protocol regarding why  BAD_FORMAT may be returned.

Thoro:
@ysmolsky Is it on purpose you added a printf for each command that is executed?

https://github.com/beanstalkd/beanstalkd/pull/562/files#diff-5b9219f6cd269a2c85dc83291ca314a4R1288

ysmolsky:
Oh wow! Thanks for spotting this. I am not sure how it got there, but it was not an intention.

tomponline:
@ysmolsky maybe we should do another release with all of the fixes you've added since the last one?

ysmolsky:
Yes, that would be awesome. Last release was almost 1 year ago...

PR #571, opened 2020-02-23
Update protocol.zh-CN.md

fantiq:
modify doc format of cmd `ignore`


kr:
Looks good, thank you for the fix!

PR #577, opened 2020-04-19
systemd: replace vendored files with libsystemd

### LINKED ISSUE #501: sd-daemon.*: replace with libsystemd on Linux
I guess this was ripped from some old systemd lib. Does it make sense to link the systemd lib on linux and just use these functions from the lib? while skipping this on other OSes?

If we have to keep it for some reason as it is, how can we test it? And what should we do with functions that are not being used?

I would prefer to get rid of it and just use the lib, but I may be missing something. @kr, can you shed light on the history of this file?

justinmayhew:
Fixes #501


README.md
```diff
@@ -29,6 +29,7 @@ also try,
     $ make check
     $ make install
     $ make install PREFIX=/usr
+    $ make SD_SOCKET_ACTIVATION=true
```
ysmolsky, line 32: This name is a little cryptic to put it into readme without explanation. Advanced systemd users might guess its meaning, but mere mortals won't.

justinmayhew: @ysmolsky agreed. Thoughts on renaming it to `SYSTEMD_SOCKET_ACTIVATION`?

ysmolsky: That is better. Please, also put a sentence explaining it.

ysmolsky:
Thanks. One thing I forgot about is the default behavior. Before this change one would not need to do anything on systemd system to enable this feature, it was enabled by default. But now one would have to set the variable. Is there a way to see if the libsystemd development files are present and enabling it in that case?

I am not the best person to decide on this. It's best answered by maintainers or providers of beanstalkd binaries on Linuxes with systemd support...

justinmayhew:
We could shell out to `pkg-config` [like Redis](https://github.com/antirez/redis/blob/c49fb47fbe0086dc9e44b8ac39128beb7ec1bf3a/src/Makefile#L149) if auto-detection is desired.

ysmolsky:
Hi. I am sorry for long wait. I hope we can continue with this. I looked into redis and I like their approach. Unless user has specified explicitly it will try to autodetect and build it with systemd support. This will keep the previous behavior intact. Maybe we can use something simple for the variable too: `USE_SYSTEMD=yes/no`?

Thank you!

justinmayhew:
Pushed an update with the auto-detection. Let me know what you think.

Makefile
```diff
@@ -43,12 +43,26 @@ TOFILES=\
...
+#        no: disabled
+#       yes: enabled, build fails if libsystemd is not found
+# otherwise: enabled if libsystemd is found
+ifneq ($(USE_SYSTEMD),no)
+ifeq ($(shell $(PKG_CONFIG) --exists libsystemd && echo $$?),0)
+	LDLIBS+=$(shell $(PKG_CONFIG) --libs libsystemd)
+	CPPFLAGS+=-DHAVE_LIBSYSTEMD
+else
+ifeq ($(USE_SYSTEMD),yes)
+$(error USE_SYSTEMD is set to "$(USE_SYSTEMD)", but $(PKG_CONFIG) cannot find libsystemd)
```
ysmolsky, line 61:
```suggestion
	$(error USE_SYSTEMD is set to "$(USE_SYSTEMD)", but $(PKG_CONFIG) cannot find libsystemd)
```

justinmayhew: Make doesn't seem to like a tab there unfortunately:

`Makefile:61: *** commands commence before first target.  Stop.`

ysmolsky: Oh, I forgot how picky the make is. Nevermind!

PR #582, opened 2020-06-01
add changelog

### LINKED ISSUE #452: Introduce change log
I propose to keep a changelog to make things easier for fellow developers/users when we release a new version. All notable changes to this project should be documented in a change log file.

I usually base mine off the "[Keep a Changelog](https://keepachangelog.com/en/1.0.0)" format. There's not really a reason to follow semver for this project, so I'd recommend just removing that part of the template.

/cc @ysmolsky @tomponline

ysmolsky:
Fixes #452

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,19 +2,15 @@
 All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
-### Added
 
-- Support of UNIX domain sockets
-- Solaris/illumos support
-- the "reserve-job" command
-- draining status to the "stats" command
-
-### Changed
-
+- add support of UNIX domain sockets
+- add support of Solaris/illumos
+- add the "reserve-job" command
+- add draining status to the "stats" command
+- make fsync turned on by default when binlog is used: it's synced every 50ms instead of never
+- replace vendored systemd files with libsystemd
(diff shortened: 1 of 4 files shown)


ysmolsky:
This is the beginning of the changelog to simplify releases. Whenever we do something important in the project we should add one line to this file. This can be used for release notes later.

CHANGELOG.md
```diff
@@ -0,0 +1,33 @@
...
+All notable changes to this project will be documented in this file.
+
+## [Unreleased]
+### Added
+- Support of UNIX domain sockets
+- Solaris/illumos support
+- the "reserve-job" command 
+- draining status to the "stats" command
+
+### Changed
```
tomponline, line 11: @ysmolsky we should probably add the default sync interval change here too.

ysmolsky: added!

ysmolsky:
I am still not sure if we need to have two section: Added/Changed. Or should we rather just copy important commits as they are without having these two sections. Right now we have this mixed, in 1.11 and 1.10 I used just copy-commit-msg style and for the unreleased I edited them by hand. I don't want to produce unnecessary friction here.

tomponline:
@ysmolsky I dont mind, what ever is easier :)

PR #583, opened 2020-06-03
server: remove redundant listen

justinmayhew:
The socket is already listening after a call to make_server_socket.


ysmolsky:
This look reasonable.

PR #588, opened 2020-07-29
Add github actions for CI

nyanshell:
Github officially provide a CI/CD feature, it's more convenient and maybe we could use it for future releases. I just added a basic make & make bench procedure. I'll try to add coverage & release later.


ysmolsky:
While I agree that Github may be better for some use cases, I fail to see why we need to migrate existing pipeline to Github.

PR #600, opened 2021-02-25
Optimize small details to make code cleaner

wuYin:
### Modification
- Removed duplicated assignment for `conn.fd`
- Reused existed `remove_ready_job` to remove job from ready heap for `process_queue`
- Enhanced `tube_find` to find in specified tube sets and not just the global tubes, use it in `OP_IGNORE` handler.

### DIFF
--- a/prot.c
+++ b/prot.c
@@ -278,6 +278,7 @@ static const char * op_names[] = {
     CMD_RESERVE_JOB,
 };
 
+static Job *remove_ready_job(Job *j);
 static Job *remove_buried_job(Job *j);
 
 // epollq_add schedules connection c in the s->conns heap, adds c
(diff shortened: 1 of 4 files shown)


prot.c
```diff
@@ -459,13 +460,7 @@ process_queue()
...
 
     while ((j = next_awaited_job(now))) {
-        heapremove(&j->tube->ready, j->heap_index);
-        ready_ct--;
-        if (j->r.pri < URGENT_THRESHOLD) {
-            global_stat.urgent_ct--;
-            j->tube->stat.urgent_ct--;
-        }
-
+        j = remove_ready_job(j);
```
ysmolsky, line 463: j could be NULL here and should be checked for that.

prot.c
```diff
@@ -1795,14 +1790,7 @@ dispatch_cmd(Conn *c)
...
 
-        t = NULL;
-        for (i = 0; i < c->watch.len; i++) {
-            t = c->watch.items[i];
-            if (strncmp(t->name, name, MAX_TUBE_NAME_LEN) == 0)
-                break;
-            t = NULL;
-        }
-
+        t = tube_find(&c->watch, name);
```
ysmolsky, line 1793: good catch.

prot.c
wuYin: Thanks for review, I added precaution check.

ysmolsky (approved):

tomponline:
thanks @ysmolsky

PR #601, opened 2021-03-09
fix default tube reference count initialized as 1

wuYin:
### Issue
Tube `default` reference counts initialized as 1, means event if no connection watching or using it, it's still in memory and never gonna be garbage collected until close server, lead to `list-tubes` / `stats` / ... are not very accurate.
```shell
> telnet 127.0.0.1 11300
use tubeX
USING tubeX
watch tubeX
WATCHING 2
ignore default
WATCHING 1
list-tubes
OK 22
---
- default  # expect deleted
- tubeX

stats
OK 1054
---
cmd-use: 1
cmd-watch: 1
cmd-ignore: 1
cmd-stats: 1
cmd-list-tubes: 1
current-tubes: 2  # expect 1
#...
```

### Modification
When new connection accepted, find or make `default` tube for it, and without macro `TUBE_ASSIGN` that has side effects, the re-make operation maybe low frequency.

### Others 
This design has the advantage of avoiding the default tube being re-make frequently, and users may not care why stats are not accurate enough. If so, it's ok to close this PR.

### DIFF
--- a/prot.c
+++ b/prot.c
@@ -2240,7 +2240,7 @@ h_accept(const int fd, const short which, Server *s)
         return;
     }
 
-    Tube* default_tube = tube_find_or_make("default");
+    Tube *default_tube = tube_find_or_make("default");
     Conn *c = make_conn(cfd, STATE_WANT_COMMAND, default_tube, default_tube);
     if (!c) {
         twarnx("make_conn() failed");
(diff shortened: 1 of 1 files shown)


ysmolsky:
Logically, default should not ever exist if no one using it. But from the point of performance, it does not make sense to ever delete it. When server crashes, it will be gone any away and in all other cases it's existence is cheaper than constant recreation. Thank for thinking about it, though.

wuYin:
Thanks for the review, make sense to me, performance should be top priority.

